PR #3, opened 2020-01-20
Request permission for takeoff

Nesiht:
Well...it takes way longer then you think to finish the project :)

I believe I could done better with the media querys and also image sizing to optimize speed when loading the site.
Also think the css could be optimized so it wouldn't end up with 600+ lines of code.

Otherwise I'am happy with the result.

Netlify:
https://clever-knuth-eee53d.netlify.com/


code/index.html
```diff
@@ -2,21 +2,169 @@
 <html>
   <head>
     <meta charset="utf-8" />
-    <title>Project 2: The News Site</title>
+    <title>The Clickbait: &#128240</title>
     <link rel="stylesheet" href="./style.css" />
-    <!-- dont forget the meta tag for responsiveness -->
+    <link href="https://fonts.googleapis.com/css?family=Alata&display=swap" rel="stylesheet">
+    <link href="https://fonts.googleapis.com/css?family=Roboto&display=swap" rel="stylesheet">
```
ericaglimsholt: Tip! You can actually select multiple fonts in Google Fonts. Then it will give you only one `<link>` to insert here. Of course this works as well, but one less row of code is great!

PR #8, opened 2020-01-21
Johannas News Site

olsenjohanna:
This project has been very good from a learning perspective, it has been a lot of trial and error but that is how I learn the best. I have realized that I can most of my answers just by googleing! (at least at this level, maybe not later ofc) 

What I'm not so proud of is the design, I'm not a designer which you can probably tell haha. 

In the end I wanted to make the images a link but doing the image resized and I did not have time look into how to fix it - hence it is only the heading that is clickable/a link. If I would had more time I would try to manage that and I would also put more time into to the header and footer. I would also have made links to real news article for all items in the grid. 

I struggled with finding out/know what override what, I added something and then something else changed.

Some questions I have:
- Using flexbox, what is the smartest/best way to set how many items/children that should be viewed per row? I use percentage of screen now, but is there any oter way? 
- I also would like to know the best way to set a space between the items/children. I would like to combine space-between and center/flex-start. Now I added a margin but what is the best way for doing this?
- If I want several (or all) anchor tags to open in a new tab, could I set that in CSS or is the only way to set it on each a-tag in HTML?

I really enjoy this, it has been a lot of fun!

(Noticed a blank line (line 46) in HTML but could not do git push again, I got an error message and did not know how what to do.)


code/style.css
```diff
@@ -1,4 +1,290 @@
...
+  margin: 0 auto;
+  padding: 50px 10px;
+}
+
+.big-news img {
+  width: 100%;
+}
+
+.big-news a:hover {
+color: #FFAEC1;
```
ericaglimsholt:
```suggestion
  color: #FFAEC1;
```

code/style.css
```diff
@@ -1,4 +1,290 @@
...
+
+.fa {
+    padding: 20px;
+    width: 20px;
+    text-align: center;
+    text-decoration: none;
+    background: #FFAEC1;
+    color: white;
+    border-radius: 20%;
+  }
```
ericaglimsholt: Unindent

code/style.css
```diff
@@ -1,4 +1,290 @@
...
+
+  .fa:hover {
+  opacity: 0.5;
+  }
```
ericaglimsholt: Unindent

PR #23, opened 2020-01-22
Done with Project 2

fridamaria:
Link to deployed version: https://ripebananas.netlify.com/

As discussed during the 2nd live session, you do not have to put the same styling in all media queries to make it work for the subsequent queries, making the code cleaner. I tried up but then it did not really work. Is this because I have a min and max width for one of the queries? Resulting in the subsequent not inheriting the same styling?

I struggled with this before adding everything back in the two queries as it seemed to work then.


code/index.html
```diff
@@ -1,22 +1,110 @@
...
+<html id="top">
+
+<head>
+  <meta charset="utf-8" />
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <title>Project 2: The News Site</title>
+  <link rel="stylesheet" href="./style.css" />
+  <link href="https://fonts.googleapis.com/css?family=Radley|Roboto&display=swap" rel="stylesheet">
+  <link rel="icon" href="banana-logo.png">
+  <!-- dont forget the meta tag for responsiveness -->
```
ericaglimsholt:
```suggestion
```

PR #24, opened 2020-01-22
Build a News Site - Lisa Sturegard

LisaStur:
Finished the project and added some hoovering effects.
Tried to add hoovering effects to the "smalll-news"-card, but could not figure out where to add the "selector in the HTML.


code/index.html
```diff
@@ -1,22 +1,153 @@
...
+
+    <h3>TIMES</h3>
+
+    <nav>
+      <a href="https://www.africaguide.com/wildlife/lion.htm" target="_blank" title="Africa Guide" class="color-hover">African Lion</a>
+      <a href="https://www.nationalgeographic.com/news/2016/08/asia-endangered-world-lion-day-conservation/"
+        target="_blank" title="National Geographic" class="color-hover">
+        Asian Lion</a>
+      <a href=https://www.panthera.org/initiative/project-leonardo target="_blank" title="Project Leonardo"
+        class="color-hover">Lion Conservation</a> </navclass="bigger-links">
```
ericaglimsholt: Hm what is going on here? Never seen `</navclass>` as an element before :)

LisaStur: Ouch, a rest of some very early (not very succesfull) attempt...  :-D 
Just glad it did not mess anything up! Big thanks for checking my code. and the stylin comment! Will kep in mind!

code/index.html
```diff
@@ -1,22 +1,153 @@
...
+  </header>
+
+  <!-- BIG NEWS SECTION -->
+  <section class="big-news">
+
+    <img src="Gapande lejon tbc.jpg" alt="Lion image">
```
ericaglimsholt: You actually don't need to have "image" in the `alt` text. Because it is an image element it will read "image" automatically. So in this case just "lion" would be enough!

PR #26, opened 2020-01-22
Project 2 - Viktoria

VickBe:
Finished project with a couple of stretch goals (Box-shadows and hover effects)
I've been struggling with GitHub!!!!! But I think I'm slowly getting there. I will be very careful to do everything by the book on the next project and commit my changes .


VickBe:
I realised when I tried to look at my site in netlify that the assets folder needed to be in the codefolder. So I had to fix that and make another pull request...

PR #27, opened 2020-01-22
Project 2 - Pål Hamel

palhamel:
I think I've finished all requirements and a few of the stretched goals. I struggled a bit with stopping at "good enough" and also in CSS to actually understand whats happening in Flex. Would love to be able to shorten the CSS "depth" if possible, to get it easier to read and understand.


code/index.html
```diff
@@ -1,22 +1,161 @@
...
+          <i class="fas fa-play-circle"></i>
+        </a>
+        
+      </section>
     </section>
-  </body>
-</html>
+  </section>
+
+<!-- FOOTER HERE, NOT SURE IF NEEDED -->
```
ericaglimsholt: A footer is always good to have :)

PR #34, opened 2020-01-23
Project 2 - Åsa Andersson

aasa72:
I planned very poorly this week, and it´s been a struggle all the way - I´ve learned that lesson well=/

What is best practice for setting height, px or auto and then padding to create space, or doesn´t it matter?
I made my footer in one <p> with <br> to create new lines, is that an acceptable way to do it or should i do several <p>? I read somewhere that you should only use lists for actual lists, so I avoided that.

I wanted to add a dropdown navbar to my hamburger menu in mobile, I found the code but I couldn´t figure out how to apply it to my existing code, so there is no menu in mobile.
/Åsa


code/style.css
```diff
@@ -1,4 +1,226 @@
...
+  nav {
+    display: none;
+  }
+  ul {
+    list-style-type: none;
+    margin: 0;
+    padding: 0;
+  }
+  img:hover {
+    opacity: 0.5;
```
ericaglimsholt: Indent one step :)

PR #38, opened 2020-01-24
Project 2 Olle Alm - Tree Times

ollealm:
Finished the site and made stretch goals with hover effects and shadows. Also tried a solution to the different image sizes and ratios issue. 

Prod of
Created a fixed ratio container that’s cropping the images using "padding-top" for height, since it can be expressed as a % of the container width. Then positioning the images on top of the padding with "position: absolute" and lastly cropping the images with "object-fit: cover."

What I would do different
Should have used more separate class names instead of just child tags like ”.big-news figure” and ”.big-news article” etc. for higher readability.

Netlify 
https://stoic-hugle-0a9beb.netlify.com/


code/style.css
```diff
@@ -1,4 +1,281 @@
...
+  color: #fff;
+  text-decoration: none;
+}
+
+a:hover {
+  color: rgb(0, 255, 0);
+}
+
+/* underlines child headlines on hover */
+a:hover > h1, a:hover > h2 {
```
ericaglimsholt: Nice one!

code/style.css
```diff
@@ -1,4 +1,281 @@
...
+  position: absolute; /* To position the img ontop of the figure padding */
+  top: 0;
+  left: 0;
+  bottom: 0;
+  right: 0;
+  height: 100%;
+  width: 100%;
+  object-fit: cover;  /* Crops the image to fit the ratio */
+
+
```
ericaglimsholt: Delete empty lines :)

code/style.css
```diff
@@ -1,4 +1,281 @@
...
+  }
+  
+  .big-news, .news-card {
+    flex-direction: row;
+  }
+
+  /* NOT NEEDED makes height of text and figure the same */
+  /* .big-news {
+    align-items: stretch; 
+  } */
```
ericaglimsholt: Delete code that you don't need :)

PR #41, opened 2020-01-24
Project 2 - Lisa

L-Lisa:
I joined the class late so have been short of time, tried to focus more on the technical flexbox practice than the design. I've been working with grids before so tried to combine those too, not sure what best practice is. Happy for any feedback. 
Finally, I had problems setting up the VS-Github-Netlify connection, any feedback on that would be useful too!
Thank you :)


ericaglimsholt:
Hi! Can you add the Netlify link to your PR or Readme? :)

L-Lisa:
Added :)

L-Lisa:
Added the meta :)

L-Lisa:
Thank you:)

PR #53, opened 2020-08-27
Urban Style - Annika Melin

ankimelin:
The most difficult thing was to understand how to get the responsive layout to have 1-2-4 cards depending on device, I went with width in the end and discoverd it was not that difficult but will look into css grid as a next step.


ankimelin:
Thanks for great and clear input, it is so nice to have some feedback on your code for future learnings! Everything makes sense and I will look into it all.

PR #57, opened 2020-08-28
News site - Claudia 

ckjerrgren:
Finished project for the News site.


code/style.css
```diff
@@ -1,4 +1,170 @@
...
+    display: flex;
+    align-items: column;
+    flex-wrap: wrap;
+    background-color: whitesmoke;
+    color: gray;
+    padding: 2rem;
+    margin: 2rem;
+    margin-right: 1rem;
+    height: 15rem;
+    width: 15rem;
```
HIPPIEKICK, line 98: If you take a look at how your page looks when it has just reached the desktop media query, you can see that the fixed width of the cards is creating a horizontal scroll. To avoid that, you can
- remove the `card-container` with flexbox and just let the grid do its thing
- remove the fixed width

This will force all the cards to only take up the space of one column. See this codepen example: https://codepen.io/hippiekick/pen/ExKwvOr?editors=1100

PR #60, opened 2020-08-28
Project 2 News site - Karolin Albinsson

karolinalbinsson:
I struggled with understanding flexbox and I'm not sure if i've used it correctly. 

I went as far as the red stretch goals. 
- Add a box shadow to make the photos look like polaroids
- Align the cards to the center of the page


EmOlsz (commented):
Overall, it's a very well done project, congratulations Karolin! You met the requirements of both blue and red level, in black level you made animations requirement, the only thing that is missing is CSS-grid. Nevertheless, here are the things you can work on : 

1. You use a lot of unnecessary line breaks between elements
2. Sometimes you miss `./` in your path that stands for `This folder`
3. In CSS, always use kebab-case naming rather than camelCase. The last one is used widely in JavaScript
4. You could change some of the `divs` to other, more semantic tags
5. Self closing tags should have ` />` at the end
6. Sometimes your code can be chaotic, it seems like you omitted last check of your code before pushing it to Github. It's your code and it stands for you, keep it clean :)

EmOlsz (approved):

PR #61, opened 2020-08-28
New site project- Ellen 

ellenbjo:
I'm aiming for the red level with this project :)

I struggled a lot with finding a good structure for the code. I experimented a lot from the beginning of the project without having a good sketch from the start and that made it harder to implement the flexbox layout.

I learned a lot from this project and had a lot of fun doing it. I'm going to read more on css grid and hopefully I can apply that to future projects. 

/Ellen


code/index.html
```diff
@@ -1,22 +1,142 @@
...
-    <section class="big-news">
-      <!-- The big news content -->
+
+<head>
+  <meta charset="utf-8" />
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <title>The News Site Project</title>
+  <link rel="stylesheet" href="./style.css" />
+  <script src="https://kit.fontawesome.com/2fb5a23775.js" crossorigin="anonymous"></script>
+  <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Cardo">
```
EmOlsz, line 10: 👍  Very nice you added custom google font

code/style.css
```diff
@@ -1,4 +1,349 @@
-html {
-  background: #0025ff;
+/**** over all styling ****/
+body{
+ background-color: #f0f0f0;
+ color: #354649;
+ margin: 0 auto;
```
EmOlsz, line 5: `auto` doesn't stand for anything over here, you can just type `margin: 0`

code/style.css
```diff
@@ -1,4 +1,349 @@
...
+  margin: 0;
+  padding: 0;
+  position: fixed;
+  top: 0;
+}
+
+.icon {
+  margin-left: 20px;
+  font-size: 48px;
+  color: #354649;;
```
EmOlsz, line 34: Some small typo - one semicolon too much

code/style.css
```diff
@@ -1,4 +1,349 @@
...
+  width: 100%;
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+}
+
+h5 {
+  font-size: 16px;
+  font-weight: bolder;
+  margin-bottom: -10px;
```
EmOlsz, line 186: Almost always there is better a solution than margin with values lower than 0. Perhaps some lower padding or margin (but still higher than 0) somewhere else?

code/style.css
```diff
@@ -1,4 +1,349 @@
...
+}
+
+.social-icons {
+  font-size: 30px;
+}
+
+/*** animation ***/
+.headline-text {
+  margin-bottom: 20px;
+  animation-name: test;
```
EmOlsz, line 201: Very nice Ellen! Maybe let's just change `test` name to something else ;)

EmOlsz (commented):
Great job Ellen! It looks like you made a lot of research regarding some CSS features. You covered requirements of both blue and red level. Things for you to keep on doing : 

1. You write clean, aesthetic code, it's a pleasure to look on it
2. From time to time you were able to nest selectors in CSS (like `.cards div { }` for example). It's good that you are aware of this CSS feature
3. You developed mobile first in a very efficient way
4. Your project is semantically correct
5. Custom hamburger menu done in CSS

Things that you can work on : 
1. Before final push, double check whether everything is named in correct way, so you can avoid `test` name of animation for example
2. You could use more classes. In CSS, we should avoid styling by selectors and in 99% of cases should do it by classes

EmOlsz (approved):

PR #65, opened 2020-08-28
Project news site 

mhammarback:
This is my code for my project. 
I have tested it on my macbook air using chrome, on my ipad and my iphone. It works as i want it to, however i am not 100% sure i used the breakpoints in the media quieries in the best way. Please let me know if i missunderstood/used it in a default way. 

I also realized when testing it on ipad/phone that the hovering effects do not work since it's touch screen.


HIPPIEKICK (changes requested):
Hi Moa! We need you to make some changes to this in order to get approved. I copy-pasted your code into a codepen to be able to explain it more easily. If you take a look at [this codepen](https://codepen.io/hippiekick/pen/JjXrQZO) first, you see that it's possible to scroll horizontally on the page - meaning that the content is wider than the page itself. When it comes to responsive web design, that's something we want to avoid. Now take a look at [this codepen](https://codepen.io/hippiekick/pen/PoNJrBL?editors=0110) with your code where I commented out all of the fixed width you've set (where you set the width of an element to be e.g. 1450px wide).

As I wrote in one of the comments, if your media query starts at 667px, your content should be able to fit within that width. This might mean that you also have to take a look at your cards section as the width (and possibly the margin and the padding) makes it so that the cards take up more width than possible.

Another thing to be aware of regarding media queries is that it's ok to have many different ones as you have (for each element), but it's more bug-prone (which means it's easier to make mistakes). Consider having all your mobile styling for different elements in one section, then a media query for all your tablet styling, and lastly one media query for all the desktop styling.

What needs to be done for this project to be approved:
- Get rid of the horizontal scroll so that your page is fully responsive (taking away the fixed width will solve the most of this issue but possibly you need to take a look at the card section as well)
- Since it's also a requirement to follow the Technigo guidelines, go through the indentation in your CSS file. It should look like this:
```
.small {
  width: 664px;
  display: flex;
  flex-wrap: wrap;
  flex-direction: column;
  margin-top: 50px;
}
```

With that said, I really like the theme, the box shadow and the hover effects on the cards - your page looks super nice - you just need to fix a little bit to get it fully responsive :) Please reach out to me if any of the comments are unclear. Feel free to tag me on Stack overflow if you have code questions.

mhammarback:
Thank you Matilda for your feedback!
I have started editing my code and i've had a few great "lightbulbs". :) 
When i am finished changing my code i will let you know!

HIPPIEKICK (approved):
Much better! Well done Moa. You still have some indentation issues in you CSS-file, where it looks like this:
```
.small {
display: flex;
flex-wrap: wrap;
flex-direction: column;
margin-top: 50px;
}
```
but should look like this:
```
.small {
  display: flex;
  flex-wrap: wrap;
  flex-direction: column;
  margin-top: 50px;
}
```

Please think about this for future projects. Approving!

PR #70, opened 2020-08-28
Handing in my news site project

HenrikeW:
I still wonder about this: https://stackoverflow.com/c/technigo/questions/1299 
Maybe you find the reason for this when you do the code review. Looking forward to it!


HIPPIEKICK (approved):
Wow, your page looks super nice and you managed well with the responsiveness! I like the box shadows and the hover effects. Your code is easy to follow and I gave some tips on how to get rid of the horizontal scroll (which we want to avoid to make it 100% responsive).

Keep up the good work Henrike!

PR #71, opened 2020-08-28
Tennigo Tennis News by Olof - Project Week 2

luuf81:
Responsive news site using flexbox.


code/index.html
```diff
@@ -1,22 +1,65 @@
...
-    <!-- dont forget the meta tag for responsiveness -->
-  </head>
-  <body>
-    <header>
-      <!-- Some kind of logo and navbar -->
-    </header>
+<head>
+  <meta charset="utf-8" />
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <title>Project 2: The News Site</title>
```
EmOlsz, line 6: Let's change title to our custom one

code/style.css
```diff
@@ -1,4 +1,188 @@
-html {
-  background: #0025ff;
-  color: #fff;
+
+a {
+  font-family: "helvetica neue";
```
EmOlsz, line 3: Let's not forget about fallback fonts

code/style.css
```diff
@@ -1,4 +1,188 @@
...
+header {
+  background: white;
+  display: flex;
+  justify-content: space-between;
+  padding-bottom: 10px;
+}
+
+header img {
+  height: 50px;
+  width: 50px;
```
EmOlsz, line 23: We should avoid setting both width and height of `img` tags, since because of that we manipulate proportions of image. If you will uncheck width proerty in devtools, you will know what I mean

code/style.css
```diff
@@ -1,4 +1,188 @@
...
+  height: 800px;
+}
+
+ /* big news article styling */
+
+.bignews-text h2 {
+  color: black;
+  text-align: left;
+    padding-left: 25px;
+    width: 100%;
```
EmOlsz, line 41: Hmm, weird indent, we can improve that

code/style.css
```diff
@@ -1,4 +1,188 @@
...
+.big-news {
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: center;
+}
+
+.bignews-img {
+  display: flex;
+  width: 50%;
+  border: black;
```
EmOlsz, line 53: Without border width and border style this property is not going to work

code/style.css
```diff
@@ -1,4 +1,188 @@
...
+
+.bignews-img img {
+  width: 500px;
+  max-width: 100%;
+  max-height: 100%;
+  height: auto;
+  margin-left: auto;
+  margin-right: auto;
+  margin-top: auto;
+  margin-bottom: auto;
```
EmOlsz, line 64: Since all of the properties are set to `auto`, we can just type one declaration `margin: auto`

code/style.css
```diff
@@ -1,4 +1,188 @@
...
+  max-height: 100%;
+  width: 50%;
+  text-align: center;
+  justify-content: center;
+  color: black;
+  font-family: "helvetica neue"
+}
+
+.bignews-text p {
+  max-height: auto;
```
EmOlsz, line 80: `auto` is not valid value for `max-height`, therefore it doesn't change anything

code/style.css
```diff
@@ -1,4 +1,188 @@
...
+
+  header nav {
+    display: flex;
+    position: absolute;
+    top: 10px;
+    left: 0px;
+    width: 100%;
+    max-width: 100%;
+    justify-content: space-between;
+    padding:0px;
```
EmOlsz, line 128: If the value is `0px`, we can omit `px` and just type `0`

code/style.css
```diff
@@ -1,4 +1,188 @@
...
+  }
+
+  header nav {
+    display: none;
+  }
+}
+
+/* animation for small article bump */
+
+@keyframes shake {
```
EmOlsz, line 176: Very nice !

EmOlsz (approved):
Good job Olof! You met requirements for blue level + animation part from black level, congratulations :) In general I can say that Your flexbox knowledge seems solid, since it's the most important part of Week 2 materials, I'm very happy to see that.  

Things you can keep on doing same way :
1. You use proper semantic elements in HTML
2. You have good understanding of flexbox

Things you can work on a bit :
1. You forgot to add alt attributes to images, let's not forget about them
2. Sometimes indents in your code are wrong, before pushing try to check if everything is ok
3. From time to time I got a feeling that you could use more classes - we should avoid styling by selectors
4. Your project meets our blue requirements in 100% which is great, but it looks a bit raw. Try to add more soul to it next time :)

PR #72, opened 2020-08-28
Project 2 - Mats Johansen

mats-johansen:
Did not get to the CSS grid in this project due to lack of time but I think I got a hold of everything else in this project. Had some trouble with the responsive design to work properly with the flexbox and hopefully I got in order now. Looking forward for the feedback on this!


code/index.html
```diff
@@ -1,22 +1,133 @@
...
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <title>SAFARISAFARIMAG</title>
     <link rel="stylesheet" href="./style.css" />
-    <!-- dont forget the meta tag for responsiveness -->
   </head>
-  <body>
+  <body>    
     <header>
-      <!-- Some kind of logo and navbar -->
+      <img class="logo" src="./images/logo2.png">
```
EmOlsz, line 11: Let's not forget about `alt` attribute

code/index.html
```diff
@@ -1,22 +1,133 @@
...
+          <a class="link" href="#about">ABOUT</a>
+        </nav>
     </header>
 
-    <section class="big-news">
       <!-- The big news content -->
-    </section>
 
-    <section class="">
+    <main>
```
EmOlsz, line 22: very nice!

code/style.css
```diff
@@ -1,4 +1,259 @@
...
+  color: #666;
+  font-size: 12px;
+  text-align: center;
+  font-weight: lighter;
+  padding: 5px;
+}
+
+.frontpage-picture {
+  max-width: 100%;
+  height: auto;
```
EmOlsz, line 106: height is `auto` by default, no need to add it

code/style.css
```diff
@@ -1,4 +1,259 @@
...
+
+.article-container img {
+  opacity: 0.6;
+  transition: opacity 1s ease-in-out;
+  -moz-transition: opacity 1s ease-in-out;
+  -webkit-transition: opacity 1s ease-in-out;
+  box-shadow: 3px 3px 1px #ccc;
+  -moz-box-shadow: 3px 3px 1px #f6efe2;
+  -webkit-box-shadow: 3px 3px 1px #f6efe2;
+  -khtml-box-shadow: 3px 3px 1px #f6efe2;
```
EmOlsz, line 170: Happy to see these vendor prefixes!

code/style.css
```diff
@@ -1,4 +1,259 @@
...
+}
+
+.article-container img:hover {
+  opacity: 1.0;
```
EmOlsz, line 174: `opacity: 1` is enough, no need to add dot

EmOlsz (approved):
Great job Mats, you met both blue and red requirements! Happy to see that :)

Here are things you can keep on doing the same way :
1. You use semantic HTML elements - great job!
2. You commented your CSS file nice way, it's easy to read thanks to it
3. I like that you used vendor prefixes for different browsers

Things you can focus on to improve :
1. Please remember about alt attribute in images
2. You could use more classes - we should avoid styling by selectors
3. From time to time it looks like you use CSS declarations you are not sure about

code/index.html
mats-johansen: Hi! Thank you for the responsen, I will take everything you said and have go through my code.

Did not know if it vendorprefixes was a good or bad idea. I actually tried to open the netlify-link on my phone yesterday (oneplus 6s) and it looked diffrent then in chromes phone-view. 

How do you style your responsive design? For example, do you change fontsize for diffrent mediaqueries? For you prefer to use vendorprefixes?

Thank you once again for the feedback / Mats

EmOlsz: Vendor prefixes was a very good idea :) Still, sometimes content of your website will looks different between browsers, it can be both because of styles and just native behaviour of how some HTML elements are displayed by browsers.

Yes, sometimes it's good idea to change font-size on different media queries. General rule is that text should be nicely visible for users.

PR #75, opened 2020-08-28
Project 2- Jessica Panditha -Newssite

jespa374:
I have tested the responsiveness of this site in the breakpoints stated in the media queries in my CSS file in Safari and Chrome. I am not so happy with the shoe image in the mobile version as it's stretched out compared to the other images. Didn't manage to fix it. I wanted to make the navbar fixed at the top but it follows along as you scroll it in the tablet version. I didn't put it in the desktop version since it didn't work entirely as I wanted to.  I would have loved to make a hamburger menu that actually opens a drop down menu but didn't succeed in it. Would have also loved make a photo gallery section with box-shadow that gives the effect of polaroids but didn't have time. Would have also anchored the photo gallery to the titles in the menu bar if I had the time.


EmOlsz (approved):
Great job Jessica! I'm impressed how experimental you are with different CSS declarations, keep on doing that :) Moreover, I like scroll-to-top button. You met both blue and red level!

Actually there are only 2 things I could tell you to improve : 
1. Pay more attention to details : sometimes you add unnecessary indents in your code, sometimes class names with capitalised letter
2. Practice spaces between elements - both br tag and margins below zero should be avoided

PR #77, opened 2020-08-28
Climate News

Corneliaryas:
I built the project mostly using flexbox, which sometimes worked as I expected and sometimes not.  I had a lot of issues with the header and navbar, mainly to separate the links on either side of the logo in tablet, yet have them all together with even space in between on desktop. I guess there is a better way than to have two "logo" elements. I did try the order property as well, but could not get that to work either. This solution was shown in my team during the demo and I do like the way it lets me control what is happening. 

I also wanted the four last small articles to have a different layout in mobile than tablet and desktop, this was very hard to figure out while doing it mobile-first, (but also coming up with the idea after built the most already) so eventually, I gave that up and added a media query for mobile. Would be nice to know if there is a way to zero- out or even change what class an element have in HTML by CSS coding.

I also will look over my naming of classes etc to the next project, now they are a bit non-consistent and messy. 

Loved to be able to add CSS animations! (As an animator my self, I of course had to try this, despite lack of time!) 
Just hope to learn when and how to use it, to not make it slow down the site too much. 

Sorry, maybe a bit too long! Could go on forever I guess!


code/style.css
```diff
@@ -1,4 +1,451 @@
...
+
+/*HEADER*/ 
+
+  header {
+    flex-direction: row;
+    justify-content: center;
+  }
+
+  .header-content {
+    width: 65%;
```
EmOlsz, line 339: You should avoid setting this value for `.header-content` - there is no reason for that, let's leave setting its width by its children. If you will check this element id devtools you will see that it's shorter than content (links) inside it

EmOlsz (approved):
Great job Cornelia, that's a well written project! You've managed to add some box-shadow and hover effect, it looks like you get flexbox as well :) 

Here are the things you should keep on doing as you are : 
1. I like that you've added custom .svg elements
2. You used semantic tags very well, I'm happy to see that
3. Moreover, your code is well commented, thanks to that it's easy to detect which part of the website is it

Things you can focus on : 
1. Remember about not setting fixed widths when it's not necessary
2. Pay attentions to values with 0 - you don't need to add any `px` there, 0 is fine enough
3. Sometimes you use camelCase way of writing classes. I'm not sure why because it's not always like that, seems like 50% of it. Please, be aware about proper, kebab-case format
4. You could work on your indents as well, sometimes they are not properly formatted

PR #79, opened 2020-08-29
Positive site

thereselinden:
**What I did:** 
Responsive news site using flexbox. 

**What I have tested** 
Chrome, firefox, safari and iphone X

**What I struggled with**
First I had a navbar where my burger was clickable and the nav ul dropped down under my header section. Which looked very good. But... when scrolling the page the nav ul was fixed on the page and covered other section when scrolling. I tried to troubleshoot that issue a lot, but gave up and created a "dummy" burger instead. I will def learn more about positioning and try this out in a later project. 

I also struggled with making my image in my first section big news to be larger than the text when changing device to e.g. desktop. Solved that by shrinking my flex-child width together with flex-grow. 

**What Im proud of** 
That I manage to style my footer with social icons. 
My hover effect on the 7 cards in the small news section.


bealun (approved):
I can tell that you've put a lot of work into your projects, it really shows! 🌟 Being just the second project and already playing around with such nice hover effects and animations is incredible. 
To push yourself a bit I would suggest to add a favicon to your site, it's a small detail but a nice one. 
Make sure you celebrate what you have accomplished this week, you should be very proud of yourself 🥳 (Obviously, you're approved!)

thereselinden:
Totally forgot to do something more about this pull request. I made another push, I hope this is enough. Thank you for the feedback =)

PR #81, opened 2020-08-29
Project 2 - Sanna Dolck Wall

sannadw:
I aimed for the blue level and tried to do some extras if I had the time.
I had some trouble figuring out flex box but the code sessions really helped me.


bealun (approved):
You are obviously very detail oriented which you will benefit from on your journey as a developer. There's a nice structure, clean code and I think you've done an incredible job with this project 🎉 A small note is that your commit messages should be in english but it's nitpicking at this point.
Make sure you celebrate this success and the fact that you've actually built this site, from scratch to an amazing news site that works on every device. CRAZY GOOD! 💃

PR #82, opened 2020-08-29
Project 2: News Site - Jamie Cook

the-j-curl:
I completed all blue requirements as well as the red stretch goals. I also completed one of the black requirements (Add some animations on your project).

I had no idea how to layout the page using CSS grid as well as including flex-box, to make the page responsive.  I found this requirement confusing. 

I used Google Chrome for testing as well as Safari for the Mobile and Tablet testing. No other browsers have been tested.


style.css
```diff
@@ -0,0 +1,322 @@
...
+
+/* Main Content */
+
+.container .main-news .main-flex-box {
+  border-bottom: 2px solid #44bfed;
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  margin-bottom: 50px;
+}
```
Cilpersson, line 66: When written without commas this styling will only affect the `.main-flex-box`. If you want it to affect all three classes it needs to written like this: 
```
.container, .main-news, .main-flex-box {
...
}
```
If you only wanted to affect the inner most container, since you have a class name for it, writing it like this would be enough: 
```
.main-flex-box {
...
}
```

Cilpersson (approved):
You've done a great job with this project and I hope that you're proud with what you've achieved. I've written some comments that you can take with you to the next project but you are definitely approved! 

Regarding the confusion you felt about grid and flexbox, here are my thoughts. 
My approach for that goal would have been to combine grid and flexbox. I would have kept the flexbox-styling for the cards you call `flex-child` and used grid to display them. So instead of using flexbox to structure the cards in the `other-news` container I would use grid. 

If you feel uncertain about things like that the Technigo SO is a great place to ask those kinds of question.

PR #86, opened 2020-08-29
Project 2 - Sara Heijkenskjöld

sarahejhej:
I tried to reach all levels. I also tried to make the code as clean and simple as possible. Would really appreciate feedback on how that worked out, if it's too simple or if other solutions would be better.


code/style.css
```diff
@@ -1,4 +1,344 @@
...
+  color: #000;
+  margin: 20px;
+  margin-right: 20px;
+  margin-left: 20px; 
+  font-size: 30px;
+}
+
+.footer a:hover {
+  color:#f8e1aa;
+  font-size: 220%;
```
EmOlsz, line 240: I just want to be sure - are you aware that `100%` of `font-size` is not `30px` in this case? `%` in `font-size` goes back to default value set by browser. In Chrome for example it's `16px`, so your `220%` will be `35px` rather than `66px`

code/style.css
```diff
@@ -1,4 +1,344 @@
...
+  }
+
+  .article-container a {
+    width: 22%;
+  }
+}
+
+/*** KEYFRAMES ***/
+
+@keyframes scale-shake {
```
EmOlsz, line 328: Nicely done!

EmOlsz (approved):
Great job Sara, I'm really impressed :) You did well with animations, box-shadow, hover effects, you even tried grid and succeed. Your flexbox and RWD skills are very good, that's the most important part of this week. Moreover, effect of skewed photos is very ambitious. I'm happy to see this kind of progress. Even though it's impressive, there is one things I would suggest you to work on : usage of classes. You seem to style a lot by selectors, think about changing that.  
Once again, congratulations!

PR #87, opened 2020-08-29
Slow Times - Hanna Skeppe

Hanna-Skeppe:


EmOlsz (approved):
Super nice project Hanna, I'm happy to see this progress! :) I like that you add meaningful comments in your code, your knowledge of flexbox and responsiveness seems to be also solid. Great job with box-shadows and experimenting with CSS properties in general!

Nevertheless, below you can find list of things you can work on a little bit more :
1. You seem to use a lot of styling be selectors - consider using classes for that
2. Some of your indents in code are wrong
3. Sometimes you use very complicated CSS code (like with links in header) that can be simplified (for example by flexbox) - take a look on that :)

PR #89, opened 2020-08-30
Project 2: Rebeka Kovačič

rebkova:


code/index.html
```diff
@@ -1,22 +1,163 @@
...
-  </head>
-  <body>
-    <header>
-      <!-- Some kind of logo and navbar -->
-    </header>
 
+<head>
+  <meta charset="utf-8" />
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <title>The New Forker</title>
```
EmOlsz, line 7: xD

code/index.html
```diff
@@ -1,22 +1,163 @@
...
+  <link rel="stylesheet" href="./style.css" />
+  <link rel="icon" type="image/png" href="./noun_fork.png"/>
+  <link href="https://fonts.googleapis.com/css2?family=Playfair+Display&display=swap" rel="stylesheet">
```
EmOlsz, line 10: Happy to see custom google font!

code/index.html
```diff
@@ -1,22 +1,163 @@
...
-    <header>
-      <!-- Some kind of logo and navbar -->
-    </header>
 
+<head>
+  <meta charset="utf-8" />
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <title>The New Forker</title>
+  <link rel="stylesheet" href="./style.css" />
+  <link rel="icon" type="image/png" href="./noun_fork.png"/>
```
EmOlsz, line 9: Great job with favicon :)

code/index.html
```diff
@@ -1,22 +1,163 @@
...
+  <link href="https://fonts.googleapis.com/css2?family=Playfair+Display&display=swap" rel="stylesheet">
+</head>
+
+<body>
+  <header>
+    <div class="hamburger">
```
EmOlsz, line 15: Custom hamburger menu, nice

code/index.html
```diff
@@ -1,22 +1,163 @@
...
+      <div></div>
+      <div></div>
+      <div></div>
+    </div>
+    <a href="#"><img class="header-image" src="./The_New_Forker_logo.png" alt="The New Forker logo" /></a>
```
EmOlsz, line 20: Since there is `img` element inside `a`, it should be present in next line

code/index.html
```diff
@@ -1,22 +1,163 @@
...
+    <nav>
+      <a href="#">Books & Culture</a>
+      <a href="#">Crossword</a>
+      <a href="#"><img class="nav-image" src="./The_New_Forker_logo.png" alt="The New Forker logo" /></a>
```
EmOlsz, line 24: Thank you for `./` in all paths!

code/index.html
```diff
@@ -1,22 +1,163 @@
...
+      <a href="#">Fiction & Poetry</a>
+    </nav>
+  </header>
+  <main>
+    <!-- The big news content -->
     <section class="big-news">
-      <!-- The big news content -->
+      <div class="big-news-card">
+        <img src="./COVID-19-adam-niescioruk-Z9arfr0f248-unsplash.jpg" 
+        alt="Black facial mask graffiti on a white wall.">
```
EmOlsz, line 34: Why this unexpected line break inside ?

code/index.html
```diff
@@ -1,22 +1,163 @@
...
+      <div class="small-news-card">
+        <img src="./troopers-jon-tyson-eIhH7RTlTZA-unsplash.jpg" 
+        alt="Image of storm troopers and Anonymous.">
+        <article>
+          <a href="#">Books & Culture</a>
+          <h2>Lorem ipsum dolor sit amet, 
+            consectetur adipiscing elit
+          </h2>
+          <p>
+            By <b><i>Ticky DeFan.</i></b>
```
EmOlsz, line 62: I'm happy to see that you know these tags, but this type of styling should be done in CSS by declarations `font-weight` and `font-style`

code/index.html
```diff
@@ -1,22 +1,163 @@
...
+        </article>
+      </div>
+      <div class="small-news-card">
+        <img src="./grafitti.jpg" 
+        alt="Colorful grafitti.">
+        <article>
+          <a href="#">Goings On</a>
+          <h2>Lorem ipsum dolor sit amet, 
+            consectetur adipiscing elit
+          </h2>
```
EmOlsz, line 86: If there is a line break between content and closing tag, it should be there also for opening tag

code/index.html
```diff
@@ -1,22 +1,163 @@
...
+          </p>
+        </article>
+      </div>
     </section>
-  </body>
-</html>
+  </main>
+  <footer>
+    <p><b>.rebkova.</b>
+      <br>The New Forker : The New Yorker magazine mock-up.
```
EmOlsz, line 159: We should avoid `br` tags as much as we can - you could wrap `.rebkova.` text inside any element with default `display: block` and result would be the same :)

code/style.css
```diff
@@ -1,4 +1,309 @@
-html {
-  background: #0025ff;
-  color: #fff;
+/** HEADER **/
+
+header {
+  padding: 10px 20px;
+  background: #FFFFFF;
+  color: #121212;
+  font-family: Helvetica;
```
EmOlsz, line 7: Let's not forget about font fallbacks

code/style.css
```diff
@@ -1,4 +1,309 @@
...
+@media (min-width: 768px) and (max-width: 1024px) {
+  header {
+    justify-content: space-evenly;
+  }
+}
+
+.hamburger {
+  display: flex;
+  flex-direction: column;
+  /*to spread the lines, since they are stacked*/
```
EmOlsz, line 22: 👍

code/style.css
```diff
@@ -1,4 +1,309 @@
...
+  justify-content: space-between;
+  height: 12px;
+  width: 18px;
+  margin: 0px, 0px, 0px, -80px;
```
EmOlsz, line 26: 1. There shouldn't be any comma between values.
2. Moreover, if you want to add value `0px`, you can just omit `px` and type only `0`. Final proper result should look like this : `margin: 0 0 0 -80px`.
3. Adding margins with value lower than zero is not the best idea, wasn't there better solution for that? Maybe you could add margin equal to `+80px` somewhere else?

code/style.css
```diff
@@ -1,4 +1,309 @@
...
+
+@media (min-width: 768px) {
+  .hamburger {
+    display: none;
+  }
+}
+
+header img {
+  width: 130px;
+  height: 30.42px;
```
EmOlsz, line 43: Most of browsers will not detect that value and just round it to `30px`, since `1px` is the smallest unit. We should write them without values after dot/comma

code/style.css
```diff
@@ -1,4 +1,309 @@
...
+  }
+}
+
+@media (min-width: 1024px) {
+  .nav-image {
+    display: none;
+  }
+  nav a:nth-last-of-type(3) {
+    display: none;
+    /*had to remove <a>; there was space left if only <img> was removed*/
```
EmOlsz, line 70: Super clever :)

code/style.css
```diff
@@ -1,4 +1,309 @@
...
+
+.big-news img {
+  width: 100%;
+  height: 200px;
+}
+
+@media (min-width: 768px) {
+  .big-news img {
+    width: 45%;
+    height: 90%;
```
EmOlsz, line 106: We shouldn't manipulate with both `width` and `height` - your image is stretched because of that

code/style.css
```diff
@@ -1,4 +1,309 @@
...
+  text-align: left;
+  color: #DB3334;
+}
+
+h1 {
+  font-family: 'Playfair Display', serif;
+  text-transform: uppercase;
+  font-size: 18px;
+  text-align: left;
+  color: #000000;
```
EmOlsz, line 167: In hexadecimal approach, if all 6 parts of a value are the same, we can just type 3 and it will be ok, so `color: #000` is completely enough

code/style.css
```diff
@@ -1,4 +1,309 @@
...
+
+/* On mouse-hover, add a deeper shadow */
+.small-news-card:hover {
+  box-shadow: 0 8px 16px 0 rgba(0, 0, 0, 0.2);
+}
+
+footer {
+  background: #000000;
+  right: 0;
+  left: 0;
```
EmOlsz, line 286: It's great you remember about that, but your footer doesn't have any `position`, so these values are not doing anything

code/style.css
```diff
@@ -1,4 +1,309 @@
...
+  padding: 10px;
+  text-align: center;
+  margin: 10px 0px 0px 0px;
+}
+
+@media (min-width: 768px) {
+  footer {
+    bottom: 0;
```
EmOlsz, line 294: Same as above

code/style.css
```diff
@@ -1,4 +1,309 @@
...
+  }
+}
+
+footer p {
+  font-family: Georgia, 'Times New Roman', Times, serif;
+  color: #ffffffa4;
```
EmOlsz, line 300: What color is that?:D Hexadecimal values should only have 6 letters (or 3, if all 6 of them are the same)

EmOlsz (approved):
Great job Rebeka, you made very nice project! Hover effects are well done, you implemented box-shadow technique, all elemenets are centered, custom favicon is added and what's most important, you understand flexbox :) 

Here are the things you can continue doing since you did a great job with them :
1. Custom google fonts
2. Comments in code when needed
3. Well usage of semantic HTML elements
4. Overall flexbox usage

Things you can work on :
1. Indentation - sometimes you forget to add new line in HTML
2. You could still read about some CSS values, you seem to not feel 100% secure with some of them

PR #90, opened 2020-08-30
Ylva's News Site

ylvan75:
Week 2 - Project - Build a News Site


code/index.html
```diff
@@ -2,21 +2,93 @@
...
       <!-- Some kind of logo and navbar -->
+      <div class="site-logo-name">
+        <img class="logo" src="./icons/icons8-sailing-ship-96.png" alt="logo" />
+        <span>SKEPP OHOJ!</span>
+      </div>
+
+      <nav>
+        <a href="#">HOME</a>
+        <a href="#">ABOUT</a>
+        <img class="logo" src="./icons/icons8-sailing-ship-96.png" alt="logo" />
```
HIPPIEKICK, line 23: You could have the image once and re-position it with `position: absolute;` instead of having the same image twice :)

code/index.html
```diff
@@ -2,21 +2,93 @@
...
+      </nav>
     </header>
 
-    <section class="big-news">
-      <!-- The big news content -->
-    </section>
+    <main>
+      <section class="big-news-container">
+        <!-- The big news content -->
+        <div class="big-news-card">
```
HIPPIEKICK, line 32: You don't need both of these as a wrapper/container, it should be enough with the `section` and then styling the image instead of the `div` wrapping it

code/index.html
```diff
@@ -2,21 +2,93 @@
...
+    <footer>
+      <p>
+        This is made by Ylva Nilsson within the project "News site" at Technigo
+      </p>
+      <a
+        target="_blank"
+        href="https://icons8.com/icons/set/sailing-ship-small--v1"
+        >Sailing Ship icon</a
+      >icon by
+      <a target="_blank" href="https://icons8.com">Icons8</a>
```
HIPPIEKICK, line 91: Be consistent (and possibly wrap the whole text or the text outside the `a` in a `p` tag):
```
<a target="_blank" href="https://icons8.com/icons/set/sailing-ship-small--v1">Sailing Ship icon</a>
<p>icon by</p>
<a target="_blank" href="https://icons8.com">Icons8</a>
```

code/style.css
```diff
@@ -1,4 +1,189 @@
-html {
-  background: #0025ff;
-  color: #fff;
+* {
+  box-sizing: border-box; /* Doing like this, is it valid for the whole page and elements? */
+}
```
HIPPIEKICK, line 3: Exactly, the `*` means that it applies to every element

code/style.css
```diff
@@ -1,4 +1,189 @@
...
+nav {
+  display: none;
+}
+
+nav a:hover {
+  font-size: 24px;
+  font-weight: bold;
+  border-top: 2px solid #056676;
+  border-bottom: 2px solid #056676;
+}
```
HIPPIEKICK, line 55: This hover styling makes all the links jump around a bit. Consider giving them a fixed width or wrap them in a div with a fixed width so that you always know how much space they will take up (to avoid the jumping around)

code/style.css
```diff
@@ -1,4 +1,189 @@
...
+  border-radius: 30px;
+  box-shadow: 5px 5px #868686;
+  width: 80%;
+  height: auto;
+}
+
+.other-news-card img {
+  width: 100%;
+  border-top-left-radius: 30px;
+  border-top-right-radius: 30px;
```
HIPPIEKICK, line 103: Could also be written:
`border-radius: 30px 30px 0 0;`

code/style.css
```diff
@@ -1,4 +1,189 @@
...
+    display: flex;
+    justify-content: space-evenly;
+    align-items: center;
+    width: 100%;
+  }
+
+  nav a {
+    color: #056676;
+    text-decoration: none;
+  }
```
HIPPIEKICK, line 134: I really like this header! 🥳

code/style.css
```diff
@@ -1,4 +1,189 @@
...
+
+  .other-news-container {
+    flex-direction: row;
+    flex-wrap: wrap;
+    justify-content: space-evenly;
+  }
+
+  .other-news-card {
+    width: 30%;
+  }
```
HIPPIEKICK, line 154: Would be nice if the cards were aligned with the big news section!

code/style.css
```diff
@@ -1,4 +1,189 @@
...
+  .big-news-card {
+    width: 30%;
+  }
+
+  .big-news-card-grow {
+    flex-grow: 2;
+  }
+
+  .other-news-container {
+    justify-content: flex-start; /* What do I need to do to get them to be center but still get the new row left aligned (not spread out)? */
```
HIPPIEKICK, line 183: `justify-content: center;` ? :)

HIPPIEKICK: Take a look at this stack overflow question as well: [https://stackoverflow.com/questions/43208025/center-a-group-of-wrapped-flex-items](https://stackoverflow.com/questions/43208025/center-a-group-of-wrapped-flex-items)

HIPPIEKICK (approved):
Good job Ylva! I really like the colors and theme of your page, it feels well suited for the topic :) I like the hover effects as well, but see if you can get rid of the jumping with help of my comment. Regarding the centering of the other news, why not write a stack overflow question about it since I think others will run into the same issue.

Keep up the good work!

PR #110, opened 2021-01-22
W. 2 - Responsive News site 

darcywinterhill:


EmOlsz (approved):
Hey Darcy!
Thank you for your project! Its looks very solid and I don't have many things to point out apart from my single comment. One extra thing I would like to mention is that on mobile devices we should avoid displaying two images (or any sections) next to each other, every single one of them should always take 100% of available space.
Since your @media parts change between resolutions correctly, I do not consider them as a reason to not accept your project, nonetheless try to remember that in future projects :)
We have prepared for all of you a document with complete set of feedback regarding ideal project for this week, check it out and try to take those tips with you to next projects!
https://www.notion.so/technigo/General-feedback-for-class-Jan-2021-ac7821ff3e1849c3bcac26f5d9888230

PR #120, opened 2021-01-22
Project week 2 - Andrea

Andrealarsson:
As a beginner in code, almost everything was a challenge but I am pleased over the results. My biggest problem was CSS, I google so much and added to much code but I have tried to make the code clean in the end. I understood git the last day so i didn't have the time to add, commit & push on this project but will be better on the next project.


code/index.html
```diff
@@ -1,22 +1,117 @@
...
+</head>
+
+<body>
+  <!-- HEADER: hamburger, title, logo & navbar -->
+  <header>
+    <div class="hamburger">
+      <div></div>
+      <div></div>
+      <div></div>
+    </div>
```
HIPPIEKICK, line 18: Nice with a hamburger 👍

code/index.html
```diff
@@ -1,22 +1,117 @@
...
+    <h1>Tasty Magazine</h1>
+    <img class="logo" src="./logo1.svg">
```
HIPPIEKICK, line 20: Don't forget the `alt` tag

code/index.html
```diff
@@ -1,22 +1,117 @@
...
+
+  <!-- BIG NEWS -->
+  <section class="big-news">
+    <div class="container">
+      <img src="images/icecream.jpg" alt="icecream image">
+      <div class="icecream">
+        <h2>Rekordår för glassbranschen under pandemiåret 2020!</h2>
+        <h3><a href="https://www.livsmedelsnyheter.se/rekordar-for-glassbranschen-under-pandemiaret-2020/"
+            target="_blank">Glassmarknaden växte med nära nio procent under pandemiåret 2020 och årets storsäljare blev
+            den svenska klassikern glassbåten...</a></h3>
```
HIPPIEKICK, line 36: New lines for nested elements:
```
<h3>
  <a></a>
</h3>
```

code/style.css
```diff
@@ -1,4 +1,310 @@
-html {
-  background: #0025ff;
-  color: #fff;
+/*MOBILE (max-width=667px) */
+
+body {
+  min-width: 500px;
```
HIPPIEKICK, line 4: What if I have a phone that is less than 500px wide? 🤔 
Skärmavbild 2021-01-29 kl  15 06 48

code/style.css
```diff
@@ -1,4 +1,310 @@
...
+}
+
+/*Big News*/
+.container {
+  background: rgb(212, 83, 93);
+  padding: 40px 0px;
+  width: auto;
+  border-radius: 5px;
+  margin: 30px;
+  min-width: 400px;
```
HIPPIEKICK, line 69: Same with this one, it causes a horizontal scroll on mobile (which is something we want to avoid when working with responsive web design). Go through your code in inspector mode (in mobile width) and check for other elements that are wider than they are supposed to.

code/style.css
```diff
@@ -1,4 +1,310 @@
...
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+}
+
+.container img {
+  border-radius: 5px;
+  width: 400px;
```
HIPPIEKICK, line 77: This for example 😇

code/style.css
```diff
@@ -1,4 +1,310 @@
...
+  body {
+    min-width: 720px;
+  }
+
+  /*Header*/
+  header {
+    width: auto;
+    background: rgba(247, 222, 0, 0.959);
+    height: 150px;
+    display: flex;
```
HIPPIEKICK, line 162: Since your first CSS code (everything written above line 151) doesn't have a media query, you can think of that as the default styling. That will apply all the time, unless you change it. Therefore you don't have to specify the same properties again, e.g. `display: flex;` is already set to the `header`, so you only have to include it in the media queries if you want to change the value of it.

code/style.css
```diff
@@ -1,4 +1,310 @@
...
+    height: 348px;
+    margin: 12px;
+    border-radius: 5px;
+    box-shadow: 5px 5px 0 0 rgba(207, 73, 84, 0.705);
+  }
+
+  .container-two img {
+    height: 348px;
+    border-radius: 5px;
+  }
```
HIPPIEKICK, line 298: Some of the images look a little distorted. Try `object-fit: cover;`to crop them and keep the correct ratio :)

HIPPIEKICK (changes requested):
Hi Andrea! Good job with this project! I like your design and the colors, gives me a summer vibe 🌺 Take a look at my comment about the width of some of your elements. If the content is too wide it will cause a horizontal scroll which is something we want to avoid when working with responsive web design, so fix that in order to be approved.

We’ve also put together some general feedback that applies to many projects this first week. Check it out and take the tips with you to upcoming projects! 👉 [General Feedback - week 2](https://www.notion.so/technigo/Week-2-Project-news-site-bbe695289d154ff98d6a247cef873f07)

code/index.html
Andrealarsson: Thanks! :)

Andrealarsson:
Thanks for your response, I have made all the changes:)

PR #124, opened 2021-01-23
News website - Ekaterina Klimenko

klimenko3007:


code/index.html
```diff
@@ -2,21 +2,163 @@
...
+              <div class="hamburger-line"></div>
+            </label>
+          </div>
+      </div>
+      <!-- Header tablet: logo in the middle, navbar on the sides-->
+      <div class="header-tablet">
+        <nav class="nav-tablet">
+          <a href="" class="nav-link">Home</a>
+          <a href="" class="nav-link">Highlights</a>
+          <img src="./Arctic-Logo.png" class="logo" alt="Logo">
```
HIPPIEKICK, line 40: The `nav` should contain only navigation elements, so the logo can be outside

code/index.html
```diff
@@ -2,21 +2,163 @@
...
+      </div>
+      <!-- Header desktop: logo on the left, navbar on the right-->
+      <div class= "header-desktop">
+          <img src="./Arctic-Logo.png" class="logo" alt="Logo">
+          <nav class="nav-desktop">
+            <a href="" class="nav-link">Home</a> |
+            <a href="" class="nav-link">Highlights</a> |
+            <a href="" class="nav-link">News</a> |
+            <a href="" class="nav-link">More</a> |
+          </nav>
```
HIPPIEKICK, line 53: Instead of having three different `nav`s, you can have only one and then style it differently with media queries. Because the content stays the same, it's just the styling that changes

code/index.html
```diff
@@ -2,21 +2,163 @@
...
+          <img class="bignews-image" src="./melting_arctic.jpg" alt="Image decription">
+          <div class="bignews-text">
+          <h2 class="bignews-headline">2020: The hottest year in the Arctic </h2>
+          <p class="bignews-para1">Temperatures in the Arctic Circle hit an all-time record in June 2020. The Arctic is warming up faster than any other region in the world.</p>
+          <p class="bignews-para2">Since the mid-1980s, the Arctic has warmed at least twice as fast as the global average, reinforcing a long downward trend in summer Arctic sea ice extent which has repercussions on the climate in mid-latitude regions.
+          </p>
+          <p class="bignews-para2">Arctic sea-ice reached its annual minimum in September, as the second lowest in the 42-year-old satellite record. Arctic sea ice for July and October 2020 was the lowest on record. Sea-ice in the Laptev Sea has been exceptionally low through the spring, summer and autumn, and the Northern Sea Route was ice-free or close to ice free from July to October 2020.
+          </p>
+          <a class="read-more" href="">Read more</a>
+          </div>
```
HIPPIEKICK, line 68: Indent nested elements

code/index.html
```diff
@@ -2,21 +2,163 @@
...
+        </div>
+      </section>
+    
+    <!-- Footer -->
+    <footer>
 
-    <section class="big-news">
-      <!-- The big news content -->
-    </section>
+      <div class="footer-general">
```
HIPPIEKICK, line 124: This could be a `<footer>`

code/style.css
```diff
@@ -1,4 +1,414 @@
...
+  margin: 0px;
+  padding: 0px;
+  font-family: Arial, Helvetica, sans-serif;
+}
+
+.container {
+  width: 100%;
+  max-width: 1200px;
+  margin: 0 auto;
+}
```
HIPPIEKICK, line 16: Would be super nice if the content of both containers were aligned:

Skärmavbild 2021-01-29 kl  16 18 40

HIPPIEKICK (changes requested):
Good job with this project Ekaterina! I like your hover effects and the different layouts for the different sections, it makes your page feel alive! However, take a look at the comments I wrote about the indentation of your HTML file. Take a look at the [code guidelines](https://www.notion.so/technigo/Guidelines-for-how-to-write-good-code-69ffb59eac4e4358a7c5752efb873c4e) if you need and change this.

We’ve also put together some general feedback that applies to many projects this first week. Check it out and take the tips with you to upcoming projects! 👉 [General Feedback - week 2](https://www.notion.so/technigo/Week-2-Project-news-site-bbe695289d154ff98d6a247cef873f07)

code/index.html
klimenko3007: I have a question. If an image is wrapped inside an <a> tag and technically a link to home page could it be part of nav? I redid my nav now without an image, but was just curious whether this is something that could be done differently in the future?

code/index.html
klimenko3007: Done :) Seemed like 3 navs was a better solution at the time :) I have a question about my dropdown menu (hamburger). Could it be repeating the text of the navs or should I somehow restyle my nav to be a part of drop down menu?

code/index.html
HIPPIEKICK: That's a good question! I guess technically, it's part of the navigation then so I would say that it's OK

code/index.html
HIPPIEKICK: If it's the same content, we should try to just style it differently with CSS instead of duplicating the HTML.

HIPPIEKICK (approved):

PR #128, opened 2021-01-23
Week 2 - Therese

ThereseHag:
I have now done my best to create a responsive News Site with a mobile first approach. 
I have experimented with hover effects, in this case for links. I have also used shadow box, but decided to not go with polaroid look but yet to try some effects to make the images more full of life. 

I also used grid, which I actually found a bit easier to deal with than flexbox. 

Almost everything was pretty new to me, since I haven't used CSS much and have never worked with responsive sites before, so all media queries were new, but I found it very useful and it actually worked quite quickly.


code/style.css
```diff
@@ -1,4 +1,327 @@
...
+
+  .grid-container {
+    grid-template-columns: 1fr 1fr;
+    grid-gap: 5px;
+  }
+
+  .shadow-1 {
+    width: 80%;
+    height: 300px;
+    object-fit: cover;
```
HIPPIEKICK, line 218: ⭐ Would be nice if all images got this!

code/index.html
```diff
@@ -2,21 +2,111 @@
 <html>
   <head>
     <meta charset="utf-8" />
+    <meta name="viewport" content="width-device-width, initial-scale-1.0" />
     <title>Project 2: The News Site</title>
```
HIPPIEKICK, line 6: You can change this to something personal!

code/index.html
```diff
@@ -2,21 +2,111 @@
...
+      </div>
     </section>
 
-    <section class="">
-      <!-- The other news  -->
+    <section class="grid-container">
+      <div class="grid-item">
+        <img class="shadow" src="football_arena.jpg" alt="Football arena">
+        <br>
+        <br>
```
HIPPIEKICK, line 54: The `<br>` element is not semantic and we should try to avoid it. You can achieve the same result using margin in this case.

code/style.css
```diff
@@ -1,4 +1,327 @@
...
+    flex-wrap: wrap;
+  }
+
+  .flex-parent {
+    display: flex;
+    flex-direction: row;
+    flex-wrap: wrap;
+    justify-content: center;
+    align-items: center;
+  }
```
HIPPIEKICK, line 168: Since this element already got some styling outside of your media queries:
```
.flex-parent {
  background: darkslategray;
  color: gold;
  margin: 5px;
  padding: 5px;
  border: black solid 4px;
  display: flex;
  flex-direction: row;
  flex-wrap: nowrap;
  justify-content: center;
  align-items: center;
  width: 100%;
  height: 300px;
}
```
you only have to specify what is different in this media query:
```
.flex-parent {
  flex-wrap: nowrap;
}
```

PR #153, opened 2021-01-24
Project 2 - Sofie Andersson

Sofie212:
I aimed for the red level goals in this project. I had a hard time grasping the Flexbox concept but eventually I think I started to get it. The responsive part was also difficult. And the way images should be sized is still a mystery to me. I tried to play around with different values and properties (borders, colors, hover effect etc) and I also managed to add some animation.

(I think) I had a problem when publishing my site on Netlify since I got a .app link and not a .com link? I spotted some question and answers regarding this i Slacks where the solution was to change the settings but I had the correct settings (I believe) so i don't think that's the issue for me.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "liveServer.settings.port": 5501
+}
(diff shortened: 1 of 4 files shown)


Sofie212:
I added the Netlify link to the Read Me section but I am not sure if it worked...

[Netlify Status](https://app.netlify.com/sites/new-site-project-two-486d32/deploys)

PR #169, opened 2021-01-25
Week-2-News-Site

TheErry:
Added all the files for my news project.


puj (changes requested):
Really cool that you decided to go with a multi-page site! Quite and undertaking at this stage!

In this particular case, I would have only created one css file.  This way I can try to re-use some of the classes on the different HTML pages.  (Each page can link to the same .css file)  This might have helped reduce the complexity of your code 👍 

Be sure to look through the General Feedback doc as well  👉 [General Feedback - week 2](https://www.notion.so/technigo/Week-2-Project-news-site-bbe695289d154ff98d6a247cef873f07)

Before we approve the assignment:
- please take into consideration the screen sizes < 360px

TheErry:
I have made said changes and suggestions :)

HIPPIEKICK:
Hi Erika! Some of your content is still too wide and causes a horizontal scroll. In mobile, I think it's the big image and on tablet, I think it's the header causing it. Play around in the inspector and have a look :)

TheErry:
It seems to have been a marginal on the other-news-card section causing the scrolling. Should be fixed now :)

HIPPIEKICK:
The margin/padding of this image still causes a little bit of horizontal scroll in small screens, so have a look at that and we're good to go 💪 

Skärmavbild 2021-04-19 kl  11 18 16

HIPPIEKICK:
Hi Erika! I saw that you made some new commits 👍  Unfortunately it seems like it didn't affect the correct media query. Also, it seems like the image ratio is off for small mobile screens. Attaching some screenshots as advice (this is how I work with the inspector, checking that it looks good on all devices, if there is room for everything inside the body or if there is content to the right of it causing horizontal scroll, what rules are applied to what media query etc). Hope this helps you, and if not - feel free to reach out 🌺 

Skärmavbild 2021-04-21 kl  09 46 16
Skärmavbild 2021-04-21 kl  09 46 53

KarolCislowski:
Hello, it's your project. Do you remember me? I may be imperfect, but I believe you can still change that before you sit down to your final masterpiece!
image

TheErry:
Hi! I probably have to book a 1:1 on this because I can't find the problem. I'm using the inspect tool, and sometimes I get a scroll and sometimes I don't. It looks like the "main" div has a margin but I can't change it in any way? Eveything looks fine for me on all screensizes otherwise.

KarolCislowski (approved):
Great site :) works great now :)

PR #174, opened 2021-03-15
Project 2, News site

AsaHildebrand:
I never did a pull request for week 2 becuase I missed that. Here it is!


KarolCislowski (changes requested):
Please add a link to the hosted version

code/style.css
```diff
@@ -1,4 +1,191 @@
 html {
-  background: #0025ff;
+  background: #d9e5ec;
   color: #fff;
+  }
```
KarolCislowski, line 4: you should reset default margins and paddings to avoid an empty frame around your page.

KarolCislowski (changes requested):
This is not a header layout that was in the provided design:
image

News cards should be the same heights:
image

All should be centralized

PR #197, opened 2021-08-28
Project News Site - The Dog Magazine

brunsant:
https://thedogmagazine.netlify.app/

Lows: I had an organized plan in the beginning, but got too much excited when I started coding and got lost in a mess of different codes. I had conflicting data everywhere and decided to start again. I also had a hard time with my footer, but it was actually a problem with another element. 

Highs: When I decided to start again, everything worked out quite easily and I am very proud that I got over that difficult time. The site looks exactly as planned at first, with a few extra touches that made it even better.


code/style.css
```diff
@@ -1,4 +1,268 @@
...
+  width: 250px;
+  border-radius: 1rem;
+  box-shadow: 5px 10px;
+}
+
+.sub-text {
+  margin-top: 0;
+}
+
+/* footer */
```
HedvigM, line 179: The footer isn't in the bottom of the page. Maybe it works to write: 
margin: 0px;
or
margin-bottom: 0px;

Or maybe it is because the footer lies inside the header in the HTML?

code/style.css
```diff
@@ -1,4 +1,268 @@
...
+}
+
+.grid-1{
+  margin-top: 20px;
+  padding: 25px;
+  align-items: right;
+  text-align: center;
+}
+
+/* medium size news */
```
HedvigM, line 125: I couldn't find any "medium sized" news on the site, many this comment is old?

code/style.css
```diff
@@ -1,4 +1,268 @@
...
+
+.small-news{
```
HedvigM, line 127: On the site it (only in desktop view) looks like the header, big news and the footer is little to "short" on the right side. I think this maybe is due to padding in the small news section.

HedvigM (approved):
Great job Bruna! It looks EPIC!!

* The code works. 
* You have met set requirements on the blue level.
* I think its a well written piece of code. It is easy to follow and you have written a nice text to the pull request. 
* I made some comments in the code, its just small things and some suggestions that may be of help. Over all it looks really good. 
*  I first opened the site on safari, it looks a bit different in some ways tex: the text in the header is between the header and "big news" element. 
* It looks like the pictures is kind of draged out in some screen sizes. 

If you have any questions, pleas write!

PR #200, opened 2021-08-28
News site project w.2 by Lisa Pousette Blomé

annaester:
https://lisapblome-w2-newssite.netlify.app/

### DIFF
--- a/code/index.html
+++ b/code/index.html
@@ -6,7 +6,7 @@
 <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <title>Project 2: birdviews</title>
     <link rel="stylesheet" href="./style.css" />
-    <link href="https://fonts.googleapis.com/css2?family=Urbanist:wght@100&display=swap" rel="stylesheet">
+    <link href="https://fonts.googleapis.com/css2?family=Urbanist:wght@300&display=swap" rel="stylesheet">
 
 
   </head>
@@ -118,6 +118,10 @@
     </div>
   
   </section>
+
+  <footer>
+    <p>All rights reserved by the bird community. <br><i>To contact us just <b>twitter</b></i></p>
+  </footer>
   
 </body>
 </html>
--- a/code/index.html
+++ b/code/index.html
@@ -4,48 +4,40 @@
     <meta charset="utf-8" />
     
 <meta name="viewport" content="width=device-width, initial-scale=1.0">
-    <title>Project 2: The News Site</title>
+    <title>Project 2: birdviews</title>
     <link rel="stylesheet" href="./style.css" />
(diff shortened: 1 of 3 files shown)


code/index.html
```diff
@@ -2,21 +2,121 @@
...
+        150 new reports of birds crashing into windows (1 553 013 this year). 
+        0 new reports of the bird flue, passing it on to hoomans seemed to work.
+      </p>
+    </div>
+
+      <div class="card-pic">
+      <p>
+      <img class="owl" src="https://pbs.twimg.com/media/EWhnmsJWoAAV_RN.jpg" alt="owl"/><br>
+      <div class="container-text">New winner of the "Long leg contest" <b><i>miss Henrietta Owl-Smith.</i></b> 
+      </p></div>
```
Skrosen, line 76: p-tag does not have to be wrapped around the img since they are all in a div which contains them. Now the img becomes the child of the p-tag, this is ofc fine but maybe you want to be able to style them individually, in that case I think you have more freedom if you keep them as siblings in the div-tag.

code/index.html
```diff
@@ -2,21 +2,121 @@
...
+        <div class="container-text">
+        <form action="https://www.rspb.org.uk/">
+        <button type="submit">Birds without borders</button></form>
+      </p></div>
+    </div>
+  
+  </section>
+
+  <footer>
+    <p>All rights reserved by the bird community. <br><i>To contact us just <a href="https://twitter.com/?lang=sv"> twitter </a></i></p>
```
Skrosen, line 118: As I was born and raised in Gothenburg - I love the fun puns a lot!

code/style.css
```diff
@@ -1,4 +1,279 @@
...
+footer {
+  border-top: 2px solid gray;
+  padding: 20px;
+  background-color: #e4eef3;
+  text-align: right;
+}
+
+/* Tablet style */
+
+@media (min-width:768px) {
```
Skrosen, line 137: Smart to style only min-width so that you don't have to repeat the same for desktop later on. Nice!!

code/style.css
```diff
@@ -1,4 +1,279 @@
...
+  .header {
+    display: grid;
+    grid-template-columns: 1fr 1fr 1fr 1fr 1fr;
+    align-items: center;
+    background-color:#e4eef3;
+    padding: 20px;
+  }
+  
+  .logo {
+    grid-column: 3;
```
Skrosen, line 152: Good thinking to get the picture in the middle of header while in tablet-mode!

Skrosen (approved):
Really fun to look at your code and GREAT JOB with your website!! 
The comments I have done on row 37 to 40 and 73 to 76 can be applied to several places in the HTML document that I think would give you more freedom to style in css as you want and would make the styling easier to play around with or change later on.

PR #208, opened 2021-08-29
Gustav Frid

gustavfrid:
Link to deployed site:
https://gustavfrid-week-2-news-site.netlify.app/


code/style.css
```diff
@@ -1,4 +1,253 @@
...
+	align-items: center;
+	text-align: center;
+	height: 100px;
+}
+footer a {
+	font-size: 0.6rem;
+}
+
+/* animation */
+@keyframes fadeOut {
```
intehon, line 174: This one is new for me! What does `@keyframes fadeOut` mean?

intehon (changes requested):
Great job Gustav! 

I really like the way you made comments in the css-file, it made it easier to understand as I wrote in the comments. Inspiring for me. 

I think it looks nice with the other news beeing as small as they are in mobile view. 

Links in header seemed unclickable at first but I later realized they where buttons without a target. Maybe it would be easier to see their function if you made the button act as a link? 

Apart from that - only changes I can see that needs to be done is to remove some unnecessary comments inte the html-file.

PR #209, opened 2021-08-29
Week 2 News Site by Lousanne

loulunds:
I am proud of the project that I did. I know that code is still messy but will improve on writing codes in the future. I realized I complicate things instead of finding simpler ways in doing things. The part that I am most proud of is the how I managed to use Flexbox to change the layout in my website though I still have some problems with the sizing and the stretching of the elements in Ipad Pro view. 

Submitted by: Lousanne Lundström

[Fake News Site](https://frosty-bhaskara-292b4c.netlify.app/)

### DIFF
--- a/code/index.html
+++ b/code/index.html
@@ -27,6 +27,8 @@
           </h2>
         </div>
       </div>
+
+      <!-- navbar -->
       <nav class="site-header-mobile-nav">
         <input id="toggle" type="checkbox" style="visibility:hidden">
           <label for="toggle">
@@ -53,7 +55,7 @@
     
     <hr class="thick" />
     <main>
-      <a id="fake_news"></a>
+      <a id="fake_news"><h2>Fake News</h2></a>
       <section id="site-big-news" class="section-news">
         <!-- big news with hero image -->
         <div class="section-news-wrap big-news-wrap">
@@ -64,7 +66,6 @@
                 alt="man in milk and cereal"
                 class="mobile-news-image hero-image"
               />
-              <div class="news-image-ovelay"></div>
             </div>
           </figure>
           <div class="site-news-body big-news">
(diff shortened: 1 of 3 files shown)


code/index.html
```diff
@@ -85,7 +86,7 @@ <h3 class="site-big-news-headline">
         </div>
       </section>
 
-      <a id="more_fake_news"></a>
+      <a id="more_fake_news"><h2>More Fake News</h2></a>
```
madeleinesvensson, line 89: The text in the cards could be next to the h3 just to clarify code a bit. Mow it says the text is a span not a headline or paragraph

madeleinesvensson, line 89: Very nice indentations which makes is easy to read! Good naming of classes and ID. It feels like there could be one less div on the cards just to make the HTML easier and shorter but I don't know the code and functions like you. Just a thought. :)

code/index.html
```diff
@@ -330,16 +331,16 @@ <h3 class="site-news-headline">
           </div>
         </div>
       </section>
+      <a id="get_help"><h2>Get Help</h2></a>
```
madeleinesvensson, line 334: maybe use - instead of _ to keep the same format on all classes and ID

madeleinesvensson (commented):
Good job with the code! The code works as it should. I could not find any big issues. The code is nicely indented and have some comments and therefor easier to understand. As commented a suggestion is to keep the text (h3/h4/p etc.) where the tag is just to make it more easy to read. As also commented it feels like there are alot of tags for the cards so maybe try to compress into fewer tags for the future? But you have also tried a lot of the tags which is good! Otherwise there is consistency in the naming and the names are desctiptive, but on the bottom there is a _ instead of - so that could be changed for more consistency. 

 The requirements for this project is met I believe! Good job on the first project! (I had a problem with commenting in the code so they might be floating above where the comment should be)

code/style.css
```diff
@@ -1,4 +1,399 @@
...
+
+* {
+  margin: 0;
+  padding: 0;
+  box-sizing: border-box;
+  width: 100%;
+}
+
+body {
+  max-width: 100%;
```
madeleinesvensson, line 11: Easy to read CSS with indentations

PR #212, opened 2021-08-29
Rebecca Philipson - Project 2

Rephili:
News site


code/style.css
```diff
@@ -1,4 +1,223 @@
...
+.other-news {
+  width: 100%;
+  height: 100vh-425px-15vh;
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+}
+
+.news-card {
```
AschwinSiaila, line 108: This looks good on your page!

AschwinSiaila (commented):
Hej Rebecca,
The page looks good! My knowledge is not enough to make any improvements for you, but I can only compare it with my experience. I made a footer with the same code as in the header and I made a flexbox around the header where i had the navbar and logo. ( I only took the logo out of the flexbox when copied it to the footer). I understood if you had more time you would do the extra requirements. But the site looks nice and understandable for me. Great job!

PR #220, opened 2021-08-29
Newsletter-project-techingo

Groovedharry:


annathunberg (approved):
The project meets all general and blue requirements. 
The site works well on all screen sizes. It's nice to look at and easy to read. Awesome job!

Very impressed that you included a toggle button using javascript, looks great!

The code is easy for me to read. The names of classes etc are descriptive and make sense. 
However, there were a few instances where I would have liked to have seen space between divs, for it to be even easier to read. I have commented on an example of such a line. 
I understand that cleaning code was not the number one priority in this project, as you also make a note of in your READme file.

For this project it's easy enough to undestand the code without comments, but I would suggest getting in the habit of commenting out the different sections, divs etc. When projects get more complex I believe this will be very helpful.

Overall a great project, I enjoyed it!

PR #225, opened 2021-08-29
New Site Project

Fatima-Gamero-Romero:
Fatima Gamero Romero


code/style.css
```diff
@@ -1,4 +1,199 @@
...
+/* HEADER STYLING */
+header {
+  background-color: whitesmoke;
+  height: 10em;
+  width: 100%;
+  position: fixed;
+  top: 0;
+}
+
+nav {
```
nama0027, line 25: wouldn't it be better to use class instead of tag for styling?

code/style.css
```diff
@@ -1,4 +1,199 @@
+/* Universal selector: it applies to all of the elements*/
```
nama0027, line 1: nice explanation!

code/style.css
```diff
@@ -1,4 +1,199 @@
...
+.cat-foto-wrapper img {
+  width: 100%;
+  height: auto;
+  border-radius: 40px;
+}
+.cat-foto-wrapper img:hover {
+  transform: scale(1.1, 1.1);
+}
+
+section.big-news h2 {
```
nama0027, line 84: wouldn't it be better to use class for h2 for styling?

code/style.css
```diff
@@ -1,4 +1,199 @@
...
+  text-align: center;
+  font-size: 1.6rem;
+  padding: 1rem;
+  color: orange;
+  text-shadow: 1px 1px rgb(46, 31, 2);
+}
+
+section.big-news p {
```
nama0027, line 92: same as previous

code/index.html
```diff
@@ -3,20 +3,127 @@
...
-    <!-- dont forget the meta tag for responsiveness -->
+    <link
+      href="https://fonts.googleapis.com/css2?family=Newsreader&display=swap"
+      rel="stylesheet"
+    />
+    <link
+      href="https://fonts.googleapis.com/css2?family=Hina+Mincho&display=swap"
+      rel="stylesheet"
+    />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
```
nama0027, line 24: i think adding comments before the actual code will increase the readability and understanding

code/index.html
```diff
@@ -3,20 +3,127 @@
...
+          Here at
+          <a href="https://www.technigo.io/program"><span>Technigo</span></a
+          >, we want the best for all our students so we have adopted cats from
+          the shelters and soon you will be receiveing them, of course, they
+          will be transported safety!
+        </p>
+      </div>
     </section>
 
-    <section class="">
```
nama0027, line 18: is there any specific reason to leave the class attribute value empty?

code/index.html
```diff
@@ -3,20 +3,127 @@
...
+        <img src="./Pictures/Cat-picture.jpg" alt="British short hair cat" />
+      </div>
+
+      <div class="big-news-text">
+        <h2>WE ARE SENDING TO ALL OF YOU A CAT!</h2>
+        <p>
+          Recently studies have shown that cat owners do have better
+          psychological health than people without pets. They help you sleeping
+          better, feeling less stressed and anxious, and even their purring
+          helps to heal tendons and muscles!<br /><br />
```
nama0027, line 55: I think its generally recommended to not use <br> tag because it is not semantic. Don't you think it would be batter to divide text in two paras using two <P> tags?

code/index.html
```diff
@@ -3,20 +3,127 @@
...
+            Announcements.</a
+          >
+        </p>
+      </div>
+
+      <div class="card">
+        <h3>TRIP TO MARS</h3>
+        <img src="./Pictures/Mars-picture.jpeg" alt="The planet called Mars" />
+        <p>
+          We are giving away 2 tickets for going to Mars! <br />
```
nama0027, line 87: same comment as i have added earlier about using <br>

nama0027 (commented):
Great job well done!. The code is working and full filling the requirements.

In summary i would suggest that considering adding class attribute for all elements and use class for styling. I think its usually consider recommended practice.  The code is well-formated and easy to read. However adding little more comments would add more charm to code readability.

code/style.css
Fatima-Gamero-Romero: Yes, sure! I tried not to use many classes as I was making only a one-page site, just to make it more simple but probably is not the best practice.

code/style.css
Fatima-Gamero-Romero: Thanks 😊

code/style.css
Fatima-Gamero-Romero: Actually yes, my code looks more complicated!

code/index.html
Fatima-Gamero-Romero: Sure!! I did not have enough time to review my code one more time, I spent so much time on flexbox that I forgot the most basic.

code/index.html
Fatima-Gamero-Romero: Not really actually! Probably I was trying something and I forgot to delete the class attribute!  Thanks for pointing this out!

Fatima-Gamero-Romero:
Great job on reviewing my code and very useful tips you have given me. Thank you so much!  😊

PR #230, opened 2021-08-29
Nina Rivera

NinaAlejandra:


IdaAspen (changes requested):
First of all, L O V E the theme! 
And your text within the Big-news section was cute and really relatable.  
Nice work with the big-news section and having the image and text responsive!  
I can see that you (like me) still figuring out the ways to get the design to work different on different screens - I was struggling a lot with that as well and got really stuck sometimes with figuring Flexbox out.... 

  Changes requested: 
  Really good that you started with mobile first! 
  - But the design doesn’t change that much from mobile to desktop right now (the Blue requirements was to have 4 columns for Desktop and 2 for tablet). 
  If you would like to continue on the design on desktop and tablet I’d loved to buddy up and share my lessons learnt from the project (I used Flexbox only)!
  
    To get the code more easy to read it can be great to add comments for guidance to make the code  easier to understand.

PR #235, opened 2021-08-30
Technigo/project-news-site/a1exa

A1eksa:


hemmahosjessi:
Aleksandra, not sure this is where I should write. ..
Great if you put a link to your Netlify site in the ReadMe file under ## View it live.

code/index.html
```diff
@@ -1,22 +1,96 @@
...
+      <h1 class="site-name">NEWS TODAY</h1>
+      <div class="navigation" id="navigation-bar">
+        <ul>
+          <li><a href="default.asp">Log In</a></li>
+          <li><a href="subscribe.asp">Subscribe</a></li>
+          <li><a href="contact.asp">Contact</a></li>
+        </ul>
+    </nav>
+  </header>
+  <hr>
```
hemmahosjessi, line 25: What is  `<hr>`  A tag for  making space? Ah, read about it now. Good to learn Thank you!

code/style.css
```diff
@@ -1,4 +1,157 @@
...
+  }
+  .big-news{
+    flex-direction: column;
+  }
+  .flex-parent{
+    grid-template-columns: repeat(2, 1fr);
+    gap: 0 20px;
+  }
 }
+@media (min-width:992px) {
```
hemmahosjessi, line 153: There is something on ipad view that makes the content align left instead, the margin on the right is much bigger. Can not fint the solution right now though.  Will try to fint it!

hemmahosjessi: Ah, it is the text that pushes all aside. So, a solution could be to change the flex-direction to column instead of row on the big-news. So the text will be under the img.  Otherwise you need to set some css to the img in smaller screens for the image to rezise down to give the text some space.

hemmahosjessi: My solution  would be to write in the mediaquery for the img in ipad view like this:
```
img.big-news-image {
    width: 50%;
}
```

code/index.html
```diff
@@ -1,22 +1,96 @@
...
+      <img src="./paper7.jpeg" alt="sas">
+      <p>Lorem ipsum is a pseudo-Latin text used in web design, typography, layout, and printing in place of English to
+        emphasise design elements over content.</p>
+      <button>read more</button>
+    </div>
+
+  </section>
+</body>
+
+</html>
```
hemmahosjessi, line 96: hmtl is really easy to read! Good structure with the indents and so. YOU ARE A PRO! :-)

code/index.html
```diff
@@ -1,22 +1,96 @@
...
+        emphasise design elements over content.Lorem ipsum is a pseudo-Latin text used in web design, typography,
+        layout, and printing in place of English to
+        emphasise design elements over content.
+      </p>
+    </div>
+  </section>
+
+  <hr>
+  <section class="flex-parent">
+    <div class="flex-child blue">
```
hemmahosjessi, line 43: Can´t find out what the colour names means in the class flex-child. There is a animation when you come in to the page that ends in white. But I see no yellow. Can you give me a hint? 

Maybe it would be good with a small description on it in the code? If it is not obvious to see on the site.

code/style.css
```diff
@@ -1,4 +1,157 @@
...
+  width: 100%;
+  
+}
+.text-container{
+  display: flex;
+  flex-direction:column;
+
+}
+
+.big-news-text{
```
hemmahosjessi, line 70: In some css styles when I click them out in the inspector it does not happen anything. An example here! 
Maybe good to go through the code and find code that is not neccessary. (I had a lot of that...) This example might be a bad one, since the img is a bit wide though. 
```
.big-news-text {
    text-align: center;
    flex-grow: 1;
}
```

code/style.css
```diff
@@ -1,4 +1,157 @@
...
-  background: #0025ff;
-  color: #fff;
+  color: black;
+  font-family: Helvetica, sans-serif, Arial,;
+  font-size: 12px;
+  background-color: white;
+  
+}
+
+.logo {
```
hemmahosjessi, line 12: Just a little tip to make the css a bit more clear to read is to be consistent in the spacing . An example:

header {
xxx
}


or

header{
xxx

}

code/style.css
hemmahosjessi: flex-direction:column;

or

flex-direction: column;

hemmahosjessi: See what I mean? :-)

hemmahosjessi (changes requested):
Really good work Aleksandra! Think that we were total newbies a couple of weeks ago. HTML really structured nice. Easy to read. The CSS could be cleaned up a bit when it comes to spacing and consistency! Just a tip to get it more readable. 

I will set this to request changes for you to fix the img in the big news on ipad view. So that the text is readable and the small news gets in the middle again.  

Tjohoo! See ya!

PR #243, opened 2021-08-31
Project News Site

intehon:
Assignment was to build a news site with different sections for The Big News and a grid of cards for other news.

The site was also requierd to be responsive and look different on different screens.

There's at least one issue left on the page to fix, but I just couldn't figure out what was wrong. The last social media icon has an orange background when hoover. 

https://intehonmag.netlify.app/


Maja473 (approved):
Obviously I'm very new to coding and code reviewing so I don't have a very keen eye, but I'm really impressed with what you've done with this and how well you managed to make it all look despite the difficulties you faced. The code looks very clean to me too and everything seems to be functioning as it should. Really good job!

PR #245, opened 2021-08-31
Jessica Falk Project New site

jessicatf:


Alisebrink:
Hi Jessica! 

I like your website, the colours and font really suits a yoga website. It contains interesting facts about the different positions etc. I have also checked out your code and compared it to the requirements for the assignment. So I have a few comments. 

** Style your page**
I can clearly see that you have styled your page. I like the coloring and the font choice. It seems suitable for a calming and informing yoga website.

** Use CSS Grid or Flexbox to layout the page **
I can see on your website and in the code that you have chosen to use Flexbox but it seems to be a bit buggy when I scale up and down your website. I can se on mobile that it works as it should but unfortunately not for iPad and desktop. 

**Create a responsive design with the following views; Desktop 4 columns, tablet 2 columns, mobile 1 column**
Since the flexbox is not working as intended the responsive design doesn't fulfill this requirements unfortunately. What you could have done different is using the attribute "gap" instead of margin for your containers withing the Flexbox-grid system. This allows you to have mutliple containers on the same row. With this you could also have used width on the image  to force it to take 50% of it's mother element. 

`gap:10px`
and
`width:50%;`

**The site should have a header, a main news article, a grid of news articles**
This one is clearly ticked off, you have a header, a main news section and a grid with several smaller news. 

**Play around with hover effects, i.e. change colors, add or remove borders or flip the image upside down.**
This one I cant really tell if you've completed or not. I atleast can't find any hover effects on your website. So unfortunately this one is not ticked off either.

PR #247, opened 2021-09-01
Mamite's project week 2 - responsive webb- mobile first

mamite100:


README.md
```diff
@@ -1,12 +1,11 @@
...
-
-## The problem
-
-Describe how you approached to problem, and what tools and techniques you used to solve it. How did you plan? What technologies did you use? If you had more time, what would be next?
+Technigo Remote Frontend Boot Camp. Our task was to create a classic magazine or newspaper site that should contain a header with a name of the site, a logo and a navbar. The site should also change depending on if you visit the site from a mobile, tablet or desktop. Further the site should have a big news section and a grid of cards with other news. On desktop sized screens, there should be four cards in a row, two in a row on tablets and on mobile, there should just be one card on each row.
+## Work 
+The sketch was made in Canva and then  the different modules were interpreted in code. 
+The site is built on html and css.
+## Result 
+The site is very basic and the mobile view is the best way to look at it.
```
elsisco, line 8: It would be great to know more about why you think the mobile view is the best. How would you have liked the tablet and desktop to display differently?

code/index.html
```diff
@@ -1,22 +1,121 @@
...
+      <div class="header-navigation-container"></div>
+      <h1 class="site-name">mamzine</h1>
+      <ul class="nav-menu">
+        <li class="nav-item" style="display:inline"> <a href="#">Link</a> </li>
+        <li class="nav-item" style="display:inline"> <a href="#">Link</a> </li>
+        <li class="nav-item" style="display:inline"> <a href="#">Link</a> </li>
+        <li class="nav-item" style="display:inline"> <a href="#">Link</a> </li>
+
+      </ul>
+      </nav>
```
elsisco, line 32: This tag I think is missing an opening tag.

code/style.css
```diff
@@ -1,4 +1,257 @@
...
+.read-more:hover{
+  background-color: rgb(137, 75, 196);
+}
+.read-more{
+  display: block;
+  padding: 12px 28px;
+  margin: auto;
+  border-radius: 5px;
+  border:steelblue;
+  cursor: poiner; F
```
elsisco, line 51: This capital "F" that has sneaked its way into the code causes some error alerts in VS Code.

code/style.css
```diff
@@ -1,4 +1,257 @@
...
+  flex-direction: row;
+  justify-content: space-evenly;
+  flex-flow: row wrap;
+  width: 100%;
+}
+.news-card-child{
+  width: 20%;
+  box-sizing: border-box;
+}
+.text{
```
elsisco, line 251: This class could have a more specific name that describes where it belongs. Maybe "big-news-text-container"?

code/index.html
```diff
@@ -1,22 +1,121 @@
...
+    <div class="news-card-child">
+      <h3>Meet Betelhem Dessie, Ethiopia’s leading tech queen</h3>
+      <img class="news-img" src="/code/img/queen3.jpg">
+      <p>I just might be a black Bill Gates in the making.I see it, I want it, I stunt, yellow-bone it
+        I dream it, I work hard, I grind 'til I own it. </p>
+    </div>
+
+    <div class="news-card-child">
+      <h3>Meet Betelhem Dessie, Ethiopia’s leading tech queen</h3>
+      <img class="news-img" src="/code/img/queen3.jpg" alt="women">
```
elsisco, line 76: Referring to an image like this seems to works with Live Server but not with Netlify.

code/index.html
```diff
@@ -1,22 +1,121 @@
...
+      <p>I just might be a black Bill Gates in the making.I see it, I want it, I stunt, yellow-bone it
+        I dream it, I work hard, I grind 'til I own it. </p>
+    </div>
+
+    <div class="news-card-child">
+      <h3>Meet Betelhem Dessie, Ethiopia’s leading tech queen</h3>
+      <img class="news-img" src="img/queen3.jpg" alt="women">
```
elsisco, line 83: This type of image reference seems to works for both Live Server and Netlify :)

elsisco (commented):
Thank you for letting me review your code! It's very interesting to see the code project of someone else. I had some thoughts apart from the smaller comments in the code. 

I think the readability in the HTML file is great. In the CSS file there is room for some improvement. If all properties are tabbed once from the left and the different styling sections are separated with a blank row browsing through the code gets easier. 

There are many semantic tags used instead of divs - thats great! I was thinking however that If a semantic tag used only once (like `<header>`) it’s not really necessary to give them a class, and I think it can even cause some confusion. The styling in the CSS file in this project both refers to the `<header>` and the class of the `.header-navigation-container` although they are the same section. Ideally there would only be one section in the CSS that styles the `<header>` to avoid contradictions. There is also a `<div>` in the header called `.header-navigation-container` but it doesn’t contain anything. If the class is removed from `<header>` and this last `<div>` gets to wrap around the links, it can be used to style them, but they are already wrapped by the `<ul>` so maybe just remove the `<div> `would be the easiest solution. 

Really good job Mamite!

PR #253, opened 2022-01-21
Idas version of project-news-site week 2

id4h4lling:
Link to my deployed project: 
https://www.notion.so/technigo/902a982c26684aad969130538bb57b40

Reflections: 
I think the CSS code could be more compact, not sure exactly how. 
And the html could maybe have better naming. You tell me! 

Also the blog posts should have links and all of the links in the nav bar too, will add when I have more time.


code/style.css
```diff
@@ -1,4 +1,309 @@
...
+
+.container-2{
+  display: flex;
+  flex-direction: column;
+}
+
+.container-2 img{
+  width: 100%;
+  height: auto; 
+  margin-inline-start: 20px;
```
HWallberg, line 169: The only thing I could find on this page that wasn't perfect was the margin for this image on mobile. With the inspector I changed this to 8px and then it seem to be centered on mobile. But 20px is perfect for the other devices.

code/style.css
```diff
@@ -1,4 +1,309 @@
...
+
+.second-news-text{
+  padding: 20px 40px 40px 50px;
+  width: 800px;
+}
+
+ .container-3{
+  display: grid;
+  grid-template-columns: repeat(4, 2fr);
+  grid-template-rows: repeat(4, 30vw);
```
HWallberg, line 307: Nice there with the grid!! If you change the repeat 4 to 2, you don't get all that empty space at the bottom on desktop.

code/style.css
```diff
@@ -1,4 +1,309 @@
...
+.menu-button {
+  display: flex;
+  height: 30px;
+  width: 30px;
+  font-size: 20px;
+  justify-content: flex-end;
+  margin: 10px;
+}
+
+/*removes button styling*/
```
HWallberg, line 51: Thanks for commenting it out! Now I know that's something you can do.

code/style.css
```diff
@@ -1,4 +1,309 @@
...
+	padding: 0;
+	font: inherit;
+	cursor: pointer;
+	outline: inherit;
+}
+
+.dropdown {
+  display: flex;
+  justify-content: flex-end;
+  align-content: flex-start;
```
HWallberg, line 65: Just by clicking around in the inspector - this doesn't do anything, or does it? ^^

HWallberg (commented):
Really impressed by the design of this page. BUT this page has more than just a just a nice look; it's also written with really good and clear code! Awesome job Ida! I've added just a few comments. I also want to add that I've learned a lot by reviewing this. You make grid not seem so scary.

PR #263, opened 2022-01-22
Project-news-site

rawisou:
**Link to my deployed website:** https://rawi-project-news-site.netlify.app/

**Reflection**
I divided the site into two parts: main news and other news.
I went with flexbox for main news and grid for other news.


code/style.css
```diff
@@ -1,4 +1,193 @@
...
+
+.other-news-polaroid {
+    padding: 10px;
+    margin-bottom: 20px;
+    background-color: white;
+    box-shadow: 0 2px 2px rgba(136, 130, 130, 0.4);
+} 
+
+/* Animation part (Demo)
+If there's no forwards, the header will disappear after it has expanded!
```
Nosslexa, line 105: Great commenting!  🥇  😊

Nosslexa (approved):
I followed these questions when looking at your code: 
​​1. Does it work or does anything look broken?
2. Am I able to understand the code easily?
3. Does the code have lots of duplicates? Could it be tidied up?
4. Does it follow the general and blue requirements?

1. The code works perfectly and so does the responsiveness. Your animation works the way you described it. 

2. I can understand your code and it's easy to follow. You have a great structure so I can follow for example where tags start and end and what elements belong to which div etc. you use the semantic tags and that also makes the code easy to follow. 
Your naming of classes is also very structured and it's easy to follow and even predict what for example an img inside a div is going to be called. 
i especially liked the commenting in your css file in the section about your animation. it was easy to follow how it's supposed to work and why you used for example forward. 
i also looked at your commit messages and even there it looked like you had a very structured way of working. 

3. Your code looks very tidy and is easy to follow, great strukture. Could not find any duplicates.  

4. You have fulfilled all blue requirements and also red and black. 

Your site looks great and so does your code. I don't have anything for you to change or correct for this project except for some more commenting in the css file like the one you did for the animation. And with this I mean mostly for future projects where I guess the css and later on JS is going to get more complex and will probably need more comments to be easy to follow. Really great jobb!

PR #268, opened 2022-01-22
Project 2: The News Site

amandatange:


code/index.html
```diff
@@ -4,19 +4,129 @@
...
+        <a class ="nav-links" href="#">Sports</a>
+        <a class ="nav-links" href="#">Technology</a>
+    </nav>
+      
+    <h1>COVID NEWS NOW</h1>
+    
+    <!-- The big news content -->
     <section class="big-news">
-      <!-- The big news content -->
+      <div class="big-news-pic"><img class="big-pic" src="https://picsum.photos/600">
```
Kras053, line 54: A suggestion here is to add a more covid-related image, and maybe some of the text could be a <p>.

code/index.html
```diff
@@ -4,19 +4,129 @@
...
+      </div>
+      <div class="big-news-text"><h2 class="biggest-heading">EU SET TO FOLLOW SWEDEN'S COVID STRATEGY: EUROPEANS RISING UP AGAINST LATEST SHUT DOWNS </h2></div>
     </section>
-
-    <section class="">
-      <!-- The other news  -->
+    
+    <section class="cards">
```
Kras053, line 59: Also here if you want to you can add more covid-related images, and change the lorem ipsum text to more article-ish text. You could also make the card into a link so it looks clickable...  or maybe the "more"-button instead, and I noticed on CNNs site that they have red hover-effect on the text if you want it to look similar.

code/style.css
```diff
@@ -1,4 +1,233 @@
...
+  text-shadow: 5px 5px rgb(194, 194, 194);
+  min-height: 4rem;
+}
+
+a {
+  color: black;
+  text-decoration: none;
+}
+
+/*Covid virus logo*/
```
Kras053, line 33: Seriously this logo wow : D

code/style.css
```diff
@@ -1,4 +1,233 @@
...
+    makes virus logo and soc-med links visible on large screens*/
+@media (min-width: 1025px) {
+  header {
+    display: flex;
+    justify-content: space-between;
+    align-items: center;
+  }
+
+  .big-news {
+    grid-template-columns: 2fr 3fr;
```
Kras053, line 202: Like you mentioned, 3 cards instead of 4, but I think it is a good initiative to go with the more stylish option!

code/style.css
```diff
@@ -1,4 +1,233 @@
...
+  font-size: min(2rem, 4vw);
+  padding: 0.5rem;
+  max-width: max-content;
+  font-weight: 500;
+}
+
+h1 {
+  font-size: 4rem;
+  text-align: center;
+  text-decoration: overline underline;
```
Kras053, line 101: so this is such a small thing so don't pay attention to it, but the lines adds up on each other on mobile media mode, though they do it better when I checked on my actual phone.

Kras053 (commented):
Hi Amanda! It was a really nice experience to do a code review on your project! You did such a wonderful and ambitious news site and should be so proud! It was very easy to read and navigate through your code, so well written! 
As far as I saw nothing was missing from the requirements, so the only thing to comment on was the hover effect on the Covid icon that you wrote about needing resizing, and the news cards missing the "more"-function in mobile mode. The only other feedback I have is the obvious that you can add more article-ish text, images and styling. 
So great job - the marquee really blew me away and also the Covid icon hover effect! Also the social media icons looked great and I want to know how you did that haha, and the more effect on the cards was also a little treat! Really inspiring project : D

PR #278, opened 2022-01-23
Week 2 - News Site

rijad90:


rijad90:
https://sleepy-volhard-fa4643.netlify.app/

code/style.css
```diff
@@ -1,4 +1,160 @@
...
+    display:block;
+  }
+
+  .header-links {
+    display: flex;
+    justify-content:space-between;
+    list-style: none;
+    cursor: pointer;
+    margin-top:130px;
+    }
```
lisabergstrom, line 125: To get the header links in the middle instead of at the far end of the right side try to decide a width for the link box.

code/style.css
```diff
@@ -1,4 +1,160 @@
...
+  height: 100%;
+}
+
+.left-header {
+  display: flex;
+  justify-content: center;
+}
+
+.right-main p{
+  font-size: 12px;
```
lisabergstrom, line 39: One tip is to use text-align: justify; if you want to make the text more even.

lisabergstrom, line 39: Have a look at the text for the mobile version. It is cut off at the edges.One solution can be to add: 
`margin-left: 50px; `
`width: 80%; `
`text-align: justify; `
But there are probably many other better solutions aswell :)

lisabergstrom (changes requested):
Great job Rijad! 
The code works goood, you have manage to create a responsive design and I really like the overall structure! It´s clean and simple! The HTML is well structured. 

One thing to think of next time can be to add some comments in your code so its easier to read it. Eg. 

Changes requested:

**For the links in the hader:**
- To get the header links in the middle instead of at the far end of the right side try to decide a width for the link box.


**For the textbox:** 
- One tip can be to use text-align: justify; if you want to make the text more even.
- Have a look at the text for the mobile version. It is cut off at the edges.One solution can be to add:
```
margin-left: 50px;
width: 80%;
text-align: justify;
```
But there are probably many other better solutions aswell :)

Overall, a really good job! 🥇

lisabergstrom:
I missed the example :)
"One thing to think of next time can be to add some comments in your code so its easier to read it. Eg. "

code/style.css
rijad90: Thank you Lisa, I actually wanted the links to be on the right side. What I did not manage to do was to a hoover effect on them though, tips are welcome for next project :)

code/style.css
rijad90: thank you for noticing this, i missed it completely! it is now adjusted to display it correctly. The problem was I had a fixed value on parent. I changed margin: 0 1em 0 1em; to margin: auto 0 1em; to display it correcly.

rijad90: thank you for noticing this, i missed it completely! it is now adjusted to display it correctly. The problem was I had a fixed value on parent. I changed margin: 0 1em 0 1em; to margin: auto 0 1em; to display it correcly.

PR #285, opened 2022-01-23
Week 2 Project News Site

Kyparn:
I did 3 pagesand i think this one what the best :)


code/style.css
```diff
@@ -1,4 +1,261 @@
...
+	.flexbox-img{
+		padding: 15px;
+		
+	}
+		/* My Grid smaller news  */
+	.grid-contanier, .flexbox-container a{
+		text-decoration: none;
+	}
+	.grid-contanier a{
+		text-decoration: none;
```
Neaa99, line 113: You could emerge this with the section above since the only element is text-decoration: none;

PR #286, opened 2022-01-23
Jessica Sandler

jessand77:
https://jessicas-news-site-project.netlify.app/


code/style.css
```diff
@@ -1,4 +1,213 @@
...
+  margin: 0;
+  background-color: #9DC88D;
+}
+
+h1, h2 {
+  font-family: 'Neuton', Garamond, serif;
+}
+
+p {
+  font-family: 'Work sans', Helvetica, sans-serif;
```
EmmaaBerg, line 20: The same font-family is added/styled for the HTML in row 6 so I think this one can be removed.

code/style.css
```diff
@@ -1,4 +1,213 @@
...
+  line-height: 1.4;
+}
+
+a {
+  font-family: 'Work sans', Helvetica, sans-serif;
```
EmmaaBerg, line 25: The same font-family is added/styled for the HTML in row 6 so I think this one can be removed as well.

code/style.css
```diff
@@ -1,4 +1,213 @@
...
+  margin: 0 auto;
+  /*Centers the main container*/
+}
+
+/* Big news section */
+
+.big-news-section {
+  display: grid;
+  grid-template-rows: 1fr auto;
+  grid-template-columns: 1fr;
```
EmmaaBerg, line 80: Is it possible to get the same result by only using column for the mobile version? In my mind I think this just gives one cell if there is one column and one row,

code/style.css
```diff
@@ -1,4 +1,213 @@
...
+  }
+
+  .header {
+    justify-content: space-between;
+  }
+
+  /*Hides the header h1*/
+  .header h1 {
+    display: none;
+  }
```
EmmaaBerg, line 169: I would like to see the name of the page for all different screen sizes. That was something I missed when looking at the page for the tablet size.

code/index.html
```diff
@@ -1,22 +1,154 @@
...
+    <header class="header">
+      <div class="logo-container"><img src="./images/carrot-icon.png" alt="carrot-icon" /></div>
+      <h1>Veggie News</h1>
+      <nav class="header-nav">
+        <ul>
+          <li><a href="#">About us</a></li>
+          <li><a href="#">Contact</a></li>
+          <li><a href="#">Nutrition</a></li>
+        </ul>
+      </nav>
```
EmmaaBerg, line 33: Good that the nav only contain links.

EmmaaBerg (commented):
Overall I thought the code had a good structure and was easy to follow. It follows the blue as well as the general requirements. I added som few comments mainly regarding the CSS.

code/style.css
jessand77: You are right, I'll remove the font-family.

code/style.css
jessand77: You are right, I'll remove the font-family.

code/style.css
jessand77: I'm not sure but I'll try to find out if it's possible.

code/style.css
jessand77: Yes, I was thinking about that too... I'll change it.

PR #290, opened 2022-01-23
Forked and committed all my news-site work into the technigo repo

JaEngd:


index.html
```diff
@@ -0,0 +1,164 @@
...
+    <div class="overlay"></div>
+
+      
+    <div class="home-text">
+        <h1>Alternative sports</h1>
+        <p>Latest news within the field of alternative sports<br>
+        Debitis veritatis voluptate officiis vero!<br></p>
+
+
+    <a href="#daily">Sport of the day</a>
```
EmmaSprings, line 59: great idea to target the links to different parts of the page, was wondering how you did that!

index.html
```diff
@@ -0,0 +1,164 @@
...
+<section class="article" id="article">
+
+
+  <h1 class="heading">Latest news in the world of <span>Alternative sports</span></h1>
+
+  <div class="box-container">
+
+        
+  <div class="box">
+            <img src="Images/Latest-8.jpeg">
```
EmmaSprings, line 77: could the naming of both folders and files be named with lowercase letters only, to not mix it up with javaScript naming? can be wrong here though!

index.html
```diff
@@ -0,0 +1,164 @@
...
+        </div>
+
+
+  </div>
+
+</section>
+
+<!--Article page ends-->
+
+<!--Sport of the day page-->
```
EmmaSprings, line 119: These comments makes it easy to navigate, great!

EmmaSprings (commented):
First of all, great work! I really like the video in the header and that you can click your way to the next section of the site, on article for example. 

When doing this review, I must admit I don't really know how to do it good since it's the first time so I'm hoping for your understanding with that. I couldn't really find anything that looks off to me. Just a reminder of the naming of the folders and files I guess, lowercase only. That's how I remember it but I could be wrong there. And it doesn't affect the site, it works as it should :) 

When checking the responsiveness, I found the padding for the article, set to 100px, made the readbility a bit hard for phone size, which could probably be fixed by changing it for the media query max 768px. When i ticked the padding off in the inspect tool, it looked great! So perhaps just needs to be padding: 0; ?

Very impressed with the hamburger menu for mobile version!! That's definitely next level! 

You ticked all the boxes for the blue level and I'm sure that hamburger menu took you right to black! Excellent job!

PR #291, opened 2022-01-23
Arianna's week 2 news site

ariallahyar:
Here is a link to my deployed site: 
https://cranky-wescoff-fb7bfc.netlify.app/


bostrombundock:
Hello! I think the code is both sheets looks very structured and clean. I might be totally wrong, but wondering if the braclets on line 38 after the @media, are the wrong way round? ( or how you say that properly in English :) )

The result, web page, looks great! Makes me hungry :)

PR #294, opened 2022-01-23
Week 2  - News site project

jenniferwvng:
Link to my deployed project: https://thirsty-varahamihira-7a8675.netlify.app

My reflections on how this week's project turned out:
Pretty good since all requirements are met and added some extras like hover etc.

Things I'd like to have clarified or explained in more detail:
None really of which I can currently think of


code/style.css
```diff
@@ -1,4 +1,244 @@
...
+  nav {
+    grid-column: span 3;
+    display: flex;
+    justify-content: center;
+  }
+
+  
+  nav a:hover, a:active {
+    color: white;
+  }
```
ariallahyar, line 230: Same as previous `nav a:hover` comment on lines 172 to 174

code/style.css
```diff
@@ -1,4 +1,244 @@
...
+  .big-news article {
+    position: relative;
+  }
+
+  nav a{
+    display: none;
+  }
+ }
+/* Mobile */
+@media only screen and (min-width: 320px) {
```
ariallahyar, line 32: This seems to apply to everything, i.e. mobile, tablet and desktop so why is the media query needed? It looks like you have the mobile styling specified starting on line 11, see below:
```css
 @media only screen and (min-width: 320px) and (max-width: 767px)

ariallahyar (commented):
Clean code! Requirements were met and code works as expected. Good use of naming conventions and good structure. Great work!

Suggestion: **commit often!**  :) 

Think of your work as “unsaved” as long as you haven’t committed. Feel free to check out my code as an example (screen clip below). I try to make a commit every time something is added, changed, or removed. It was definitely hard to remember at first.
Screenshot 2022-01-26 at 16 22 34

code/style.css
jenniferwvng: True! Thanks a lot for pointing it out, I figured how to do it for w.4 project :)

jenniferwvng:
> Clean code! Requirements were met and code works as expected. Good use of naming conventions and good structure. Great work!
> 
> Suggestion: **commit often!** :)
> 
> Think of your work as “unsaved” as long as you haven’t committed. Feel free to check out my code as an example (screen clip below). I try to make a commit every time I feel like I've achieved something (added, changed, or removed). It was definitely hard to remember at first. Screenshot 2022-01-26 at 16 31 17

Thanks a lot for this suggestion, it improved my workflow for the next project even though I still have to practice it! :)

PR #295, opened 2022-01-23
Sofia Ringstedt project week 2

sofiaringstedt:
Same description as in readme!

# News Site

This project has been the very first webpage I have done! The goal was to build to work with either flexbox or grid and media queries to create a repsonsive webpage. 

## The problem

Since this was the first time for me, my goal was to create a website with a proper build up and structure. I think this was my biggest challenge besides from proper css selection. I started working with my desired design in figma and also mobile-first. If I had more time I would try to better understand the proper strucuture of HTML, how to select in CSS in a cleaner way and also how hiearchy works. I struggled a lot with the image cropping and the flex box of the bigger articles. I would have wanted to not mix viewport and %, and I would also have designed the big article differently (I tried, and failed with row-reverse)

## View it live
https://hungry-easley-29ee74.netlify.app/


code/style.css
```diff
@@ -1,4 +1,207 @@
...
+  list-style-type: none;
+  margin: 0;
+  padding: 0;
+  display: flex;
+  justify-content: space-evenly;
+}
+
+nav {
+  width: 40%;
+}
```
jenniferwvng, line 56: The comment /**/ you wrote describing what html section these CSS styles rule made it clear and also simplified readability, nice! However I believe writing ```nav ul {}``` could increase quick readability even more, also in case if there were to be other ul in the page.

code/index.html
```diff
@@ -2,21 +2,139 @@
...
+
+<footer>
+  <nav>
+    <ul>
+     <li>Vision</li>
+     <li>Contact</li>
+     <li>Subscribe</li>
+   </ul>
+  </nav>
+  <div class="Name">
```
jenniferwvng, line 131: I believe you just didn't see this minor thing before hand-in since you formatted your code great elsewhere, but the tab indents should be aligned

code/style.css
```diff
@@ -1,4 +1,207 @@
...
+/*These style the footer OBS! navbar styling in header*/
+
+footer {
+  display: flex;
+  align-items: center;
+  justify-content: space-evenly;
+}
+
+
+@media (max-width: 667px) {
```
jenniferwvng, line 155: I found this article https://www.freecodecamp.org/news/taking-the-right-approach-to-responsive-web-design/ regarding mobile-first design, where they encourage usage of min-width - and I see that both are in use in the code, not 100% sure but think it's encouraged to use min-width from mobile breakpoints and counting upwards to desktop breakpoints.

jenniferwvng (commented):
Really nice code, clean structuring, cool functioning website meeting all requirements, cool with the animation! No further comments, good job!

code/style.css
sofiaringstedt: Agree! I really tried to implement this but did not manage without destroying the flex. Will continue to try!

PR #296, opened 2022-01-23
project news site, Daily Space Kids Edition

EmmaSprings:
Link to deployed site: 

https://reverent-kirch-ef3988.netlify.app/

Working on this project has been a ride, let's say that. It's been fun, frustrating, annoying, exciting and time consuming. 
This is also my first time doing a pull request so this probably wont be perfect either. :)

After trying grid at first, I decided to go for flex box for the layout as it seemed a bit easier to grasp. I haven't grasped it entirely yet but I feel I'm getting there (at some point). 

My issues have been for example, cropping the images into a size, I'm not happy with them being different sizes but can live with that for now. And getting my flex children in rows of 3 or 4. Still don't understand how to decide on that. 

I also struggled with the responsiveness and did from desktop view to mobile, which wasn't preferred but thats how i started. Next time I will do it the other way around. 

I'm also learning how to structure my code so this project won't be that pretty but hopefully can be a bit understood by the viewer. 

Overall happy with my final result, it's been a ride for sure! 

Please tell me what you think, be honest, I can handle it :)


sofiaringstedt:
Please forgive that I put all the comments here and not directly into the code. For some reason it wont allow me to comment on your code that is just just the recent changes in the last commit, and I also cannot see a html file in you latest commit. If this could be an error with your commit, perhaps you could try and re-do it and I could give your comments into the code. (I think it could be something with git add .  perhaps? 
Here goes: 

Overall a very nice looking webpage, love how nice you made the header look and your solution with the upper navbar in mobile version (that it goes in a column). Just something that caught my eye is that you have many rows of css (around 300), perhaps could be an idea to look at your webpage in inspect - tick off and on the boxes and see if there are any unnecessary css :) 

With the logo image: If you hover now you can see how the div class logo-type is bigger than the image itself in mobile mode. I wonder if setting a width on the div class logo-type (as big as you want the image inside it to be), and then simply put the logo.img to 100% ( so it adjust to its parent, the div logo-typ). I think this could also help the logo not to be cut off in certain iPad modes (but be responsive all the way) 

I see you wrote in your readme about trouble with image sizes, I had the same!! My solution was honestly to just make sure the images were the same size, I used https://croppola.com/. I am sure there is a better way tho…. Perhaps again with sizing the parent correctly and then putting the img to 100% of it? Perhaps content-overflow…? 

It seems that you have a little bit of margin to the right of the page. I am not (!!) certain why this is but perhaps it could be an idea to put a <main> to wrap the page (except header) and then putting a set width to it? 

Putting a <main> could (maybe) also center your smaller news article a bit more; if you put to it (it=the main)
	.	    display: flex;
	.	     flex-wrap: wrap;
	.	     justify-content: center;
}

Overall a very nice looking and functioning website, great design and also easy to read! I love your hover effects, the expand and the inverted pictures. Def. meets the requirements. Great job!

PR #298, opened 2022-01-23
Project 2-Coffee news site 

TereseClaesson:


code/style.css
```diff
@@ -1,4 +1,139 @@
...
+}
+
+.footer {
+  background: black;
+  text-align: center;
+  font-size: 1rem;
+  padding: 0.5rem;
+}
+
+@media (min-width:0px) and (max-width:667px){
```
joannaringqvist, line 124: I think that just `@media (max-width:667px)` is enough here. You don't need the min-width: 0px;

joannaringqvist (approved):
Super well done! I think your code was very clean and well structured! 👍 🥳

PR #301, opened 2022-01-23
Li Lindblom

motherofbeasts:


code/style.css
```diff
@@ -1,4 +1,229 @@
...
+   .news-card {
+   height: 200px;
+   width: 200px;
+   margin: auto;
+   margin-top: 20px;
+   flex-direction: column;
+   flex-basis: 14%;
+   }
+
+}
```
michaelchangdk, line 228: SUMMARIZING COMMENTS:

Overall, super impressed by your site - I do remember Friday you told us you hadn't begun yet, and you have fulfilled all the minimum requirements laid out in the brief.

Some things that I noticed in the site, but which I wasn't sure where to comment:
There was some inconsistency using margins and paddings, both in values, but also a good mix of % and pixels. I can see why - you were trying to align parts your site. I think in the future, it would be easier sticking to one or the other. Another tip is to use the same left-right padding on the section containers - that way you don't have to add it to the actual elements inside the site, and it makes it easier to see what's going on, and to keep things aligned.

HTML also automatically adds padding to things like h1, h2, h3, h4, p, etc - in the beginning of the CSS document, it can be useful to remove all default padding by setting it to 0, so you can control it better yourself in the rest of the document.

I can recommend a Chrome extension called "Pesticide" - when you activate it, you can see all the boxes and margins/padding of how the site is built up. It makes it a lot easier to inspect when using the inspect tool in Chrome and see what's going on "behind the scenes", since all our sites are made of boxes.

Overall I think you used flex really well, and your content was a fun read! Awesome job!

❤️ Michael

michaelchangdk (commented):
Left my final comments at the bottom of the CSS document. Great job! 😄 
❤️ Michael

PR #302, opened 2022-01-24
week-2-news-site

mathildakarlsson:
Link to the website: https://objective-tereshkova-645921.netlify.app

I realized that I've barely made any comments in the code.. If I would do it over again I would definitely add comments! 

I'm glad that the small javascript-project worked, which was the open and close menu in mobile version - when the nav-links can be viewed by clicking the "hamburger-menu".

One thing that I struggled with was to make a good responsive nav-bar. The links kind of merges in with the logo when you make the screen smaller, just before it hits the breakpoint and changes to tablet or mobile. I tried to make each nav-link-child take up as little room as possible but I'm not completely satisfied..

I like the hover-effect when the images zooms in a little, but I can't make it as smooth as I want it. The images kind of moves a little up to the right when it "zooms in", which is especially noticeable in desktop..

Oh, and I'm sorry for the weird name, "test", for the final push. I had a lot of struggles with git the last couple of hours and when it finally worked I didn't want to risk anything by doing another push to change the "name" 😅  I hope I will learn all this to next week's deploy!


amandatange:
First off, great job! Your website looks really good and there are some really nice style choices on there like for example the hovering effects and the color changing background. 

You mention the navbar as an area where you had some trouble and I think I agree with you. If you look at line 60 in your css file 

".nav-links li {
  padding: 1.5rem;
}"

I think you could change the padding to 0 or maybe like a 0.5vw which would allow it to shrink a little an allow every link to fit before that breaking point where the hamburger menu appears.

Beyond that, I think maybe it would improve the readability a little to rename some of the divs regarding the navbar. I'm mostly thinking about the div "navbar" on line 15 in your html which could maybe be mistaken with the <nav>. Perhaps that 'navbar' div could be changed for a <header> as well. 

The navbar looks really good and I think the hover effects especially are done really nicely with the links getting both bigger and changing color!

PR #303, opened 2022-01-24
Final hand in week 2

isabellwastfelt:


isabellwastfelt:
https://inspiring-bhaskara-9a841b.netlify.app/code/

code/style.css
```diff
@@ -1,4 +1,275 @@
...
+  text-decoration: none;
+  font-weight: bold;
+  margin-left: 20px;
+}
+
+a:hover {
+  background: gray;
+  transition: .5s;
+  border-radius: 20px;
+}
```
JensBergqvist, line 57: The links are beautiful! I really like the effect you have when you hover over them!

code/style.css
```diff
@@ -1,4 +1,275 @@
...
+
+.logo-hidden {
+  display: none;
+}
+
+
+/* ----- Big news ----- */
+.big-news {
+  display: grid;
+  grid-template-columns: 50% 50%;
```
JensBergqvist, line 73: Maybe try just 1 column with 100% so everything is in 1 row for mobile.

JensBergqvist (commented):
Great structure in the code overall! I know where I am what does what at all times!
You made a beautiful site and should be really proud of it!

I only saw 1 thing that might need a change In mobile verision. in the "Big news"-section The info about the new magazine should be below the image instead of next to it? like instead of having grid-template-columns: 50% 50%. maybe try with  :100%? I think it was on row 73 in the Css file.

PR #390, opened 2023-01-16
Eva Sjöholm - Pre Course Project

EvaSjoholm:


EvaSjoholm:
https://astonishing-flan-bb7e56.netlify.app/

EvaSjoholm:
I found this project for a total beginners stage to be a bit to major in all honesty.
My experience from the get-go was with a lot of hiccups that I needed extended help for to fix regarding downloads, the terminal, GitHub access etc. 

The biggest challenge with the Project was grasping the navbar solution at first. Through google search and W3 Schools I found a skelett that I chose to go by. Then I didn't fully understand how Flex and Grid could work together, didn't find a red thread about these two through all the tutorials, websites, YT or anywhere. A friend within IT explained it and showed me, then it made more sense because I could see first before actually just doing and not understanding why it isn't working. These type of questions was the hardest to not being able to discuss with someone who know's at first. This goes for the media-part to code as well. 

I am proud of the outcome even if it took me 60+ hours.
My stubborn and disciplined personality trait was very well spent and also that I do dare to ask for help and how much I really do want to learn, I seize the challenge and I loved the creative part of it!

PR #447, opened 2023-08-07
Unicorn News

saradahlstrom:


saradahlstrom:
Okay, looking on it from a mobile I can tell it's not as responsive as I thought when using a virtual copy of a mobile phone in a browser... Lesson learned to double check :)

PR #460, opened 2023-08-13
newsproject

EmmyLJ:
This is the best I can do at the moment. Just wanted to hand the pre-course in before the intro days so you can see where I am right now in my learning process.


EmmyLJ:
Try to hand it in again.

PR #468, opened 2024-01-17
Gossip Gazette - News Site

ohitsnathalie:


HIPPIEKICK (approved):
Wow Nathalie, the grid looks so good! Really cool that you implemented a hamburger menu as well. The requirement says that it should be 4 columns on desktop, but it looks so good with 3 so I'll allow it.

It feels like you put a lot of thought into your design choices in general, so kudos for that.

One thing that I would like you to think about is to keep your code clean - to always keep the CSS to the CSS file (you styled images in the HTML). This will make the code easier to read and maintain. If you're unsure of how to write clean code, please have a look at the resource from the pre-course material.

The CSS file could also be "cleaned up" a bit, with consistent indentations and not nesting elements.

Overall, you've done a good job so keep it up :)

ohitsnathalie:
@HIPPIEKICK Thank you for your feedback. I will keep in mind to have a cleaner CSS.
May I ask, what screen size you have? 😅
There should be 4 columns from a screen width of 1200px. At least I can see 4 columns on my 13" monitor.

PR #471, opened 2024-01-23
Puppy News- project-news-site

ericamechler:
Link to my deployed version, which could also be found in the readme-document:

https://puppy-news.netlify.app/


HIPPIEKICK (approved):
Good job with the pre-course Erica! I really like the hamburger menu (just that the white text on light blue background is a bit hard to read). I _love_ that you kept the margins consistent, so that the "big news" align with the smaller ones, it makes it look so neat.

You implemented flex and grid in a nice way, and not to be forgotten: that puppy is so cute!

PR #473, opened 2024-01-24
project-news-site-linda-frischknecht

linda-f:
Please find attached my project from the Pre-course.
I had a lot of fun learning more about HTML & CSS.
However there where some problems to solve along the way. 
But I think it turned out quite all right for my first real project.


HIPPIEKICK (approved):
I really like that you implemented a hamburger menu! It looks a little bit buggy when it opens though, that the elements jump around a bit you know?

Some tips:
- When you want an anchor tag to look like a button - you can style it to look like a button. No need to wrap the link in a button element.
- For elements that should look the same, we can give them the same class names (we reuse them). This will make our CSS file less repetitive when we can target only one class name. When all the grid areas are the same size, we don't need to name them differently, we can work with fraction instead.

The scrolling banner is a really nice addition, and your page is responsive and following the requirements ⭐️

PR #478, opened 2024-01-27
Travel Today

Antontall:


HIPPIEKICK (changes requested):
Good job with the pre-course, and good implementation of CSS Grid. You haven't used flexbox anywhere, and since that is a requirement I will request a change before approving this.

Another thing I would like you to have a look at is how your page looks like in 320px width, the images are a bit uneven and the content is not centered anymore. I suppose that's not intentional?
Skärmavbild 2024-01-31 kl  11 08 50

It seems to be the same with tablet, that the last image is a bit too big. Apart from that, I think you managed well with the responsiveness. I really like that you used consistent margins for your news-site ⭐️

PR #485, opened 2024-01-28
Erika Wernbro: Veggie News Site

Tejpex:
My veggie news site: https://warm-panda-028459.netlify.app/


HIPPIEKICK (approved):
Good job with the pre-course Erika!
I would just like to give some small suggestions:
- It would look so good if you'd center the content.
- Also consider resizing the images a bit in mobile, because they grow quite long on big mobile screens.
- Another small thing that would look good would be to make the caption as big as the images:
Skärmavbild 2024-01-31 kl  13 07 48
- Also, have a look in the inspector to see how your page looks like in 1600px width. I think a good idea could be to set a max-width on your content to create some space on the sides and make the lines a bit shorter.

Apart from these things, I think it looks good and ou managed well with flexbox and grid. I like that your media queries are short and clean ✨

PR #487, opened 2024-01-28
Precourse project

viktoria-olo:
I had to redo the pull request. This is the new version.


HIPPIEKICK (approved):
Good job with the pre-course Viktor! Your grid looks good on all screen-widths, the only thing I would suggest is to use a bit more of the width on big desktop - so the text on these small cards don't get so cramped:
Skärmavbild 2024-01-31 kl  11 19 07

I really like the box shadow of the cards, and the font decision is on point 😉 Keep up the good work!

PR #520, opened 2024-08-24
The Featherington – A news site about birds

HeleneWestrin:
Here's my finished project.


JennieDalgren (approved):
Great work completing your news site! I love the bird theme 🐦 
You've met the requirements, here comes some feedback:

### CSS Layout Simplification: 
You've mixed Grid and Flexbox effectively, but you might want to simplify in areas where one layout method can do the job. For instance, in the grid of articles, you are using display: flex; with grid properties, which can be redundant.

Example:
```
display: grid;
grid-template-columns: repeat(4, 1fr);
gap: 20px;
```
You could apply this uniformly without needing flex properties like justify-content.

### Media Queries: 
Nice use of media queries for responsiveness! However, consider using more standard breakpoints to ensure wider compatibility with various devices. You’ve set custom breakpoints like 900px; switching to widely used breakpoints like 768px for tablets and 320px for mobile would be helpful.

Example:
```
@media (max-width: 768px) {
  /* Tablet styles */
}
```


### Accessibility: 
You have multiple `<h1>` tags on the page. It's best to have only one `<h1>` per page for SEO and accessibility reasons. You can use other headings like `<h2>`, `<h3>`, etc., for the remaining sections.
### CSS Cleanup: 
Review and remove any unused or redundant CSS rules. This will make the code cleaner and more efficient.

HeleneWestrin:
Thank you so much for your feedback, @JennieDalgren! 🙏  Much appriciated.

Regarding your example CSS Grid code, where exactly do you mean I should use something like that instead? Is it on ".grid-wrapper" that is the container of my articles?

Also I can't seem to find the multiple <h1> tags you mentioned? 🤔

JennieDalgren:
My bad regarding the `<h1>` - there's only one, just as it should be! 💪

PR #521, opened 2024-08-25
Main

xingyin2024:
Week 2 | Project assignment

This is a news site about Architecture design and image which is built by Xing Yin and updated 20240825.


xingyin2024:
Here is my website link
https://xingsnews.netlify.app/

JennieDalgren (approved):
Great job with your News project. For next time, remember to add your Netlify link in the Readme file too. Here are some feedback:

### Responsive Design: 
You’ve incorporated media queries well to adjust the layout for mobile and tablet. One suggestion would be to use a min-width approach rather than max-width for media queries. This makes it easier to handle future breakpoints and device ranges.

Example:
```
@media (min-width: 768px) {
  .news-grid {
    grid-template-columns: repeat(2, 1fr);
  }
}
```

### Redundant Code: 
There are a few places where the CSS could be cleaned up. For instance, some properties are repeated or unused. Take the time to go through the styles and remove any redundancy to keep your CSS lean and efficient.

Example:
```
.header {
  font-size: 20px;
  color: black;
}
.header {
  margin-bottom: 20px;
}
```
These two declarations could be combined into one block.

In general, your work is off to a great start! Keep up the good work. 😊

PR #534, opened 2024-08-25
Project-news-site-Indian-food

IamLise:
On this website you will find the best Indian restaurants in Stockholm!


JennieDalgren (changes requested):
Great job with this project!

The layout effectively adapts across different screen sizes, but on mobile screens, it doesn’t consistently show a single column as expected. The grid structure might still display multiple columns on smaller devices.

To fix:
- Make sure the grid shows only 1 column on mobile view:

image

PR #539, opened 2024-09-02
Week 2: Project New site

jacquelinekellyhunt:
Here's my PR for the Project "News Site"


JennieDalgren (approved):
Great job with the project!

Nice job with the hamburger menu and the nice animation to go from hamburger to cross. However, it looks a bit broken on small mobiles.
image



There is too much padding on the grid when in mobile size and also on the tablet size. Try to use as much of the screen that you can! 
See the difference here:
image

And then with that huge padding:
image

I will approve this project but at least make sure to change so that the hamburger menu doesnt look broken on smallest screens

Keep up the good work! (Also love the surf theme 🏄🏼‍♀️)

PR #543, opened 2024-10-06
Accesibility Project - Elina Eriksson Hult

ElinaEH:
https://67028c384213c4000827a34d--effulgent-empanada-6d5956.netlify.app/


HIPPIEKICK (approved):
I'm glad you managed with the image in the end, I think this looks super! And I want to tell you, I can totally relate to wanting to see how it looks deployed (I have totally commited changes just to trigger a new deploy 😅

Now, to accessibility!
- Lighthouse report succeeded
- You have useful alt tags
- Contrasts are OK

Some suggestions for improvements:
- Why a button inside an `<a>`? 👀 Can't see when that would be neccessary
- Consider adding aria labels to improve accessibility for screen readers

Nothing you need to change now, just something to think about going forward!

PR #544, opened 2024-10-06
Vinify access version

smily342:


HIPPIEKICK (approved):
Good job with accessibility! Alt text descriptions for images are appropriately provided, making them accessible to screen readers, and all your contrasts are sufficient ⭐️ A suggestion for improvement could be to look into ARIA labels and roles. This was not a requirement for this week, but something that you might want to look into going forward for better screen reader support 😊

